PR #4, opened 2019-03-06
Use cobra, Unmarshal() configuration, Use knownhosts ssh subpackage

arapov:
See https://github.com/spf13/cobra/blob/master/README.md for details.


cmd/root.go
```diff
@@ -0,0 +1,78 @@
...
+		viper.AddConfigPath(home)
+		viper.SetConfigName(".cpma")
+	}
+
+	viper.AutomaticEnv() // read in environment variables that match
+
+	// If a config file is found, read it in.
+	if err := viper.ReadInConfig(); err == nil {
+		fmt.Println("Using config file:", viper.ConfigFileUsed())
+	}
```
gildub: I believe we need to trap the error case by reversing the test:
'''go
	if err := viper.ReadInConfig(); err != nil {
		log.Fatalf("%s", err)
	}
	fmt.Println("Using config file:", viper.ConfigFileUsed())
'''

arapov: Yeah, updated. I will update fmt. bits to log. tomorrow morning so it is streamlined across sources.

arapov:
> I'm not keen to encapsulate the ssh/sft into the configuration.
> We will do it but more likely as a separate internal package.
I guess we need to chat about this. I don't understand this comment.

> Also the list of files to be retrieved is going to be provided from a scan on the cluster entry point.
> I think you previous commit was good. And we can continue from there.
> What do you think?
Works for me.

PR #5, opened 2019-03-08
Add -o --output-dir option

arapov:
this is helpful for
- defining the output directory for extracted control plane
  configuration in general
- as well as to run the tool in the loop against various target
  control planes and save the output in a respective directories


arapov:
for instance this makes possible to run something like this in shell:

for i in configs/*.yaml; do
	go run ../cpma.go --config $i -o $(basename ${i%.*})
done;

PR #8, opened 2019-04-01
Adds interface for top structure and Fetch, Show methods

gildub:


env/env.go
```diff
@@ -16,14 +19,42 @@ type Clusters map[string]NodeConfig
...
+
+type Cmd interface {
+	FetchSrc() int
+	Show() int
 }
 
 type Info struct {
-	Cluster    Clusters        `mapstructure:"cluster"`
+	SrCluster  Clusters        `mapstructure:"srcluster"`
+	DsCluster  Clusters        `mapstructure:"dscluster"`
```
arapov: @gildub if you don't expect these to be [de]serialized with [un]marshal() you don't want to mark those with `mapstrcuture:"..."`.

PR #11, opened 2019-04-02
set the default outputDir and make it OS agnostic

arapov:
Current dir is the default for the output.


arapov:
it's tiny => bypass review.

PR #21, opened 2019-04-10
Add custom logger

alexander-demichev:
It might be better to have custom logger defined in a separate package. 
Log package is a wrapper around logrus logger, it's a popular library that has more functionality than standard log. This wrapper should initialize logger on first call and output all logs to both stdout and cpma.log.

Moreover, this PR introduces `--debug` command, that sets log level to debug, allowing us to print any dev info we want.


internal/sftpclient/sftpclient.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	"io"
 	"io/ioutil"
-	"log"
 	"os"
 	"path"
 	"path/filepath"
 	"time"
 
 	"github.com/fusor/cpma/pkg/config"
+	"github.com/fusor/cpma/pkg/log"
```
arapov: @alexander-demichev I didn't notice you did this for config too... IIRC, pkg/* is the code which is Ok to use in external projects. This is not the case. We have internal/* for this.

go.mod
```diff
@@ -25,7 +25,7 @@ require (
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/pkg/profile v1.3.0 // indirect
 	github.com/pkg/sftp v1.10.0
-	github.com/sirupsen/logrus v1.4.1 // indirect
```
arapov: What actually // indirect means? That gomod thing figured the dependency on its own?
I need to learn how to update/upgrade deps later. :)

alexander-demichev: Yes, go mod did this automatically :)

Issue #26, opened 2019-04-11
cpma falls FATAL when run with --help flag


arapov:
```
$ ./bin/cpma --help
Helps migration cluster configuration of a OCP 3.x cluster to OCP 4.x

Usage:
  cpma [flags]

Flags:
      --config string       config file (default is $HOME/.cpma.yaml)
      --debug               set log level to debug
  -h, --help                help for cpma
  -l, --local-only          do not fetch files, use only local files.
  -o, --output-dir string   set the directory to store extracted configuration.
40/40 10:40:41 FATAL unable to read private key: open : no such file or directory
```
That's wrong, right?

PR #28, opened 2019-04-11
Fix error on --help command

alexander-demichev:
Moving logic away from main() should fix error we see when using `--help`. In future I would prefer to remove code from  `rootCmd.Run`


alexander-demichev:
fixes https://github.com/fusor/cpma/issues/26

Issue #25, opened 2019-04-11
cpma should be able to get fqdn/ip or local file/dir without exlicitely specifying the source


arapov:
ssia. E.g.:
$ cpma some.ocp3.paas.host.com
vs
$ cpma ./this/dir/here

and get rid of -l/-local-only flag

arapov:
Pull request #35 somewhat addresses it. Though we want to think deeper on CLU UX to decide what exactly should be the user interaction.

PR #35, opened 2019-04-14
Oauth refactor

arapov:
1. Once we know more about CRDs, and OAuth settings for OCPv4 specifically, we want to generalize OCP4 Oauth code.

2. Use example configuration yaml to run app.
2.1. Use hostname as a *Source* in yaml for the first time.
2.2. Replace hostname in *Source* by local directory, so you don't wait for SSH everytime.

This is robust prototype, feel free to improve.


ocp4crd/oauth/oauth.go
```diff
@@ -0,0 +1,135 @@
...
+	configv1 "github.com/openshift/api/legacyconfig/v1"
+	oauthv1 "github.com/openshift/api/oauth/v1"
+	"gopkg.in/yaml.v2"
+
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func init() {
+	configv1.InstallLegacy(scheme.Scheme)
```
arapov: This one is not needed here. Though let it be here for now and remove in future commits.

ocp4crd/oauth/oauth.go
```diff
@@ -0,0 +1,135 @@
...
+// Generate converts OCPv3 OAuth to OCPv4 OAuth Custom Resources
+func Generate(masterconfig configv1.MasterConfig) error {
+	var auth = masterconfig.OAuthConfig
+	var err error
+
+	var crd v4OAuthCRD
+	crd.APIVersion = "config.openshift.io/v1"
+	crd.Kind = "OAuth"
+	crd.MetaData.Name = "cluster"
+
```
arapov: This stuff above, I guess, is possible to decipher from the original master-config and get rid of hard coding.

examples/cpma-config.example.yaml
```diff
@@ -1,6 +1,6 @@
 ---
-Source:
-  UserName: "root"
-  HostName: "master-0.gildub1.lab.pnq2.cee.redhat.com"
-  SSHKey: "/home/gildub/.ssh/quicklab"
-outputPath: "./data"
\ No newline at end of file
+Source: "master-0.gildub1.lab.pnq2.cee.redhat.com" # can also point to directory
```
arapov: After the first run you can make Source equal to OutputDir and CPMA will be working with the local configuration files copy.

env/env.go
```diff
@@ -1,334 +1,17 @@
...
-	"encoding/json"
-	"fmt"
-	"html/template"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"reflect"
-	"strings"
-
+	"github.com/davecgh/go-spew/spew"
```
arapov: spew is temporary. this is TODO: to figure out good logger which can pretty print structures.

cmd/root.go
```diff
@@ -48,13 +48,9 @@ func Execute() {
 func init() {
 	cobra.OnInitialize(config.InitConfig)
 	rootCmd.PersistentFlags().StringVar(&config.ConfigFile, "config", "", "config file (default is $HOME/.cpma.yaml)")
-	rootCmd.PersistentFlags().BoolVar(&debugLogLevel, "debug", false, "set log level to debug")
+	rootCmd.PersistentFlags().BoolVar(&debugLogLevel, "debug", false, "show debug ouput")
```
arapov: help, more useful for end user

cmd/root.go
```diff
@@ -48,13 +48,9 @@ func Execute() {
...
 
 	rootCmd.Flags().StringP("output-dir", "o", "", "set the directory to store extracted configuration.")
-	config.Config().BindPFlag("outputPath", rootCmd.Flags().Lookup("output-dir"))
-	config.Config().SetDefault("outputPath", path.Dir(""))
-
-	rootCmd.Flags().BoolP("local-only", "l", false, "do not fetch files, use only local files.")
```
arapov: we don't want to abuse CPMA with parameters. See the comment to example-config below.

ocp4crd/oauth/oauth.go
```diff
@@ -0,0 +1,135 @@
...
+			idP.Challenge = p.UseAsChallenger
+			idP.Login = p.UseAsLogin
+			idP.MappingMethod = p.MappingMethod
+			idP.GitHub.HostName = github.Hostname
+			idP.GitHub.CA.Name = github.CA
+			idP.GitHub.ClientID = github.ClientID
+			idP.GitHub.Organizations = github.Organizations
+			idP.GitHub.Teams = github.Teams
+			// TODO: Learn how to handle secrets
+			idP.GitHub.ClientSecret.Name = github.ClientSecret.Value
```
arapov: Secrets can be in different places: ENV var, configmap, secrets, plaintext. We have to learn how to handle those.

ocp4crd/oauth/oauth.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+	}
+
+	yamlBytes, err := yaml.Marshal(&crd)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Println(string(yamlBytes))
+	return err
```
arapov: This printout CRD yaml. We rather need to implement proper file writers. Also, I'm not sure if we want to split various providers into multiple CR files.

alexander-demichev:
Something seems to be broken here:
```
ademicev-OSX:cpma ademicev$ ./bin/cpma --config /Users/ademicev/Documents/go/src/github.com/fusor/cpma/examples/cpma-config.yaml
36/40 10:36:52 FATAL unable to parse configuration: 1 error(s) decoding:

* 'Source' expected type 'string', got unconvertible type 'map[string]interface {}'
```

ocp4crd/oauth/oauth.go
gildub: Yes that's the plan.

alexander-demichev:
Do we want to move `ocp3` and `ocp4` packages under `internal` dir?

alexander-demichev:
> Do we want to move `ocp3` and `ocp4` packages under `internal` dir?

Works fine, didn't notice you changed config

cmd/root.go
gildub: +1

ocp4crd/oauth/oauth.go
gildub: The way OCP4 works is by components, so we're likely to break it down the same way.

arapov:
This implements #25

PR #37, opened 2019-04-15
Remove config package

alexander-demichev:
This PR removes config package and moves main() logic to cmd, so we could avoid hitting error when using `--help`.

I decided to drop all structures related to config and keep just the object provided by Viper library. In my opinion this may be more flexible, because we are not limited by Info struct and have unique object for storing all config related data.
Moreover, minor change on client initialization. Let's init it at the beginning and use as package var.

Please, share your thoughts. I'm open to any improvements.


cmd/root.go
```diff
@@ -46,11 +60,11 @@ func Execute() {
 }
 
 func init() {
-	cobra.OnInitialize(config.InitConfig)
-	rootCmd.PersistentFlags().StringVar(&config.ConfigFile, "config", "", "config file (default is $HOME/.cpma.yaml)")
+	cobra.OnInitialize()
+	rootCmd.PersistentFlags().StringVar(&env.ConfigFile, "config", "", "config file (default is $HOME/.cpma.yaml)")
```
arapov: We don't need default config place. I think it's wrong for this tool and what user would expect. Wdyt?

cmd/root.go
```diff
@@ -34,6 +37,17 @@ var rootCmd = &cobra.Command{
...
+
+		// TODO: Passing *e.Info here is not exactly nice. Fix?
+		ocp3config := ocp3.New()
+		ocp3config.Fetch()
+
+		m := ocp3config.ParseMaster()
+
+		oauth.Generate(m)
 	},
 }
```
arapov: Ack

env/env.go
```diff
@@ -1,27 +1,48 @@
...
+	}
+	viperConfig.Set("home", home)
+
+	if ConfigFile != "" {
+		// Use config file from the flag.
+		viperConfig.SetConfigFile(ConfigFile)
+	} else {
+		// Search config in home directory with name ".cpma" (without extension).
+		viperConfig.AddConfigPath(home)
+		viperConfig.SetConfigName(".cpma")
```
arapov: ...

env/env.go
```diff
@@ -1,27 +1,48 @@
...
+	}
+
+	viper.AutomaticEnv() // read in environment variables that match
```
arapov: Cool stuff, I don't know

internal/sftpclient/sftpclient.go
```diff
@@ -54,19 +60,19 @@ func (c *Info) NewClient(hostname string) (*Client, error) {
 	}
 
 	// TODO: accept custom port
-	addr := fmt.Sprintf("%s:22", hostname)
+	addr := fmt.Sprintf("%s:22", source)
```
arapov: We want to make sure we can set custom port here. Not in this patch though.

cmd/root.go
alexander-demichev: @gildub What do think about this?

arapov: It's not a content for this patch. Feel free to merge.

PR #38, opened 2019-04-15
Moves cmd execute to init

gildub:


gildub:
After that one, I'd be keen to move back some of the `&cobra.Command` to main.

PR #41, opened 2019-04-16
Added an example master-config with github oauth integration

Danil-Grigorev:
Added master config tested on OCP3.7 with github OAuth2 integration for future testing.


arapov:
@alexander-demichev, have a look, this is one more example of github oauth from OCPv3.7. We have to decide, whether we want to test against all the OCPv3 minor versions. And if we do - we want to figure out how to keep this in project structure.
CC: @gildub

gildub:
@alexander-demichev, we might encounter similar difference across the board between ocp 3 versions. For instance node-config, etc. Therefore we could store the difference under `example/ocp-3.<x>` or something equivalent), for instance `example/ocp-3.7/etc/origin/.../`

alexander-demichev:
@gildub I agree, we should store it under `ocp-3.x`directories.

arapov:
@alexander-demichev, is it good to go?

PR #43, opened 2019-04-17
fixups

arapov:
ssia 

### DIFF
--- a/internal/sftpclient/sftpclient.go
+++ b/internal/sftpclient/sftpclient.go
@@ -12,9 +12,10 @@ import (
 
 	"github.com/fusor/cpma/env"
 	"github.com/pkg/sftp"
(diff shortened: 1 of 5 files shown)


internal/sftpclient/sftpclient.go
```diff
@@ -52,15 +53,13 @@ func NewClient() Client {
...
 
-	port := sshCreds["port"]
-
-	log.Debugln("Using port: ", port)
-
-	if port == "" {
-		port = "22"
+	port := 22
+	if p := sshCreds["port"]; p != "" {
+		port, _ = strconv.Atoi(p)
```
alexander-demichev: Should we return error if atoi fails?

arapov: Resolved with commit bb794bf

PR #46, opened 2019-04-17
Clean-ups and updates to ocp[3,4]/*

arapov:
- oauth: pass OAuthConfig as parameter to Convert
- rename ocp3config to ocp3


gildub:
@arapov, ah names choice! ;)
I agree `Generate` was to be replaced, I already applied `Translate`, although `Convert` is a good candidate. Let's @alexander-demichev vote! 
Otherwise LGTM

arapov:
> Please, rebase as I moved oauth and so into ocp4.

Rebased. I will look into your code from last evening during today.

PR #50, opened 2019-04-18
Streamline logger, Update TODO/FIXUPs, Small CodeFixes.

arapov:
ssia

### DIFF
--- a/ocp4/oauth/oauth.go
+++ b/ocp4/oauth/oauth.go
@@ -16,10 +16,6 @@ func init() {
 	oauthv1.Install(scheme.Scheme)
 }
 
-// TODO: Generated yamls are results of pure imagination. Structure and consistency
-// must be reviewed and fixed. I guess this code can be simplified once we know
-// how the output should look exactly.
-
 // reference:
 //   [v3] OCPv3:
 //   - [1] https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#identity_providers_master_config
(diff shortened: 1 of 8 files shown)


ocp4/oauth/oauth.go
```diff
@@ -16,35 +16,26 @@ func init() {
...
-
 // Shared CRD part, present in all types of OAuth CRDs
 type OAuthCRD struct {
-	APIVersion string   `yaml:"apiVersion"`
-	Kind       string   `yaml:"kind"`
-	MetaData   Metadata `yaml:"metaData"`
-	Spec       struct {
+	APIVersion string `yaml:"apiVersion"`
+	Kind       string `yaml:"kind"`
+	MetaData   struct {
```
gildub: The reason I used a separate type for Metadata is because it seems to be common practice but also we need to be able to reuse it for Secret among other things, actually Metadata is every where (or so until (and if) we have OCP4 structure from k8s libs).

ocp4/secrets/secrets.go
```diff
@@ -1,23 +1,25 @@
...
+	ClientSecret string `yaml:"clientSecret"`
 }
 
 type Secret struct {
-	ApiVersion string      `yaml:"apiVersion"`
+	APIVersion string      `yaml:"apiVersion"`
 	Kind       string      `yaml:"kind"`
-	Type       string      `yam:"type"`
+	Type       string      `yaml:"type"`
 	MetaData   Metadata    `yaml:"metaData"`
```
gildub: Here is the other Metadata part.

PR #51, opened 2019-04-18
Add GitLab oauth provider

alexander-demichev:


ocp4/oauth/oauth.go
```diff
@@ -68,6 +68,10 @@ func Translate(oauthconfig *configv1.OAuthConfig) (*OAuthCRD, []secrets.Secret,
 			idP, secret := buildGitHubIP(serializer, p)
 			oauthCrd.Spec.IdentityProviders = append(oauthCrd.Spec.IdentityProviders, idP)
 			secrets = append(secrets, secret)
+		case "GitLabIdentityProvider":
+			idP, secret := buildGitLabIP(serializer, p)
+			oauthCrd.Spec.IdentityProviders = append(oauthCrd.Spec.IdentityProviders, idP)
```
alexander-demichev: I don't like this part, we repeat same lines 3 times in each `case` statement. But I didn't find a way to avoid this in golang

arapov: It is easy: common part within the switch + goto! :rofl:

arapov:
Looks good.

PR #57, opened 2019-04-18
[WIP]Retrieve file on demand

gildub:
That work but wrong design which is shown in test setup blocker as we have to mock the file retrieval.
A better design is to have the communication module creating a queue per host and process file fetch when a new element is in the queue.


gildub:
@arapov, I'm still trying to figure out what's the best way to mock `ocp3.FetchFile` func.

gildub:
This approach is not good and is going to be replaced using a pipeline workflow

PR #59, opened 2019-04-23
Add keystone oauth, secrets refactor and tests

alexander-demichev:
I made secret generation more flexible and covered it with tests.

Keystone auth misses file fetching, same as htpasswd now.


ocp4/secrets/secrets.go
```diff
@@ -45,6 +51,21 @@ func GenSecretFile(name string, encodedSecret string, namespace string) *Secret
 	return &secret
 }
 
+func buildData(idenityProviderType, encodedSecret string) interface{} {
```
alexander-demichev: I believe this way of building secret data could be reusable for all oauth providers

gildub:
@alexander-demichev, thanks for that. Well, I believe we can do even be more generic with the secrets but let's be pragmatic for now ;)

PR #60, opened 2019-04-23
Move common attributes to a separate struct

alexander-demichev:


gildub:
@alexander-demichev, thanks.

PR #64, opened 2019-04-23
Pipeline workflow

gildub:
Implement go routines to provide pipelines which takes a file at the entry point and provide manifest(s) at the other end

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -21,12 +21,18 @@ import (
 	"path"
 	"path/filepath"
 
+	//"github.com/fusor/network"
+
 	"github.com/fusor/cpma/env"
 	"github.com/fusor/cpma/ocp3"
 	"github.com/fusor/cpma/ocp4"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+type File struct {
+	URL string
+}
 
 func init() {
 	cobra.OnInitialize()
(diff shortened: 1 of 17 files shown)


dymurray:
It looks like the intention of this PR is to allow the fetching/translation of files to happen in parallel to reduce computation time. Is that accurate?

arapov:
This thing is evil. Looks more crazy than the loop we have in https://github.com/fusor/cpma/blob/master/cmd/root.go#L58 now, which I have no idea why...

gildub:
@dymurray, the computation gain is a side effect, the main gain is the workflow. The second commit shows the whole in action.

gildub:
@arapov, independently the pipeline itself, the loop event can be removed, this was just the idea of detecting changes on the disk and reiterate until times is up.

eriknelson:
> (the idea was to detect changes on the disk, submit a newly found file to the pipe and reiterate until times is up.).

Is this actually a requirement for the project? My understanding of the problem, in a nutshell, is we need fetch N files, applyl a transform to each file, and export the OCP4 assets to disk for the admin to configure their OCP4 cluster at their leisure. Please correct me if that's an oversimplification, but if true, I don't see a need to be watching or responding to events. It's an [ETL tool](https://en.wikipedia.org/wiki/Extract,_transform,_load).

If it *is* a requirement to start this and watch for changes, then an event loop feels like a more correct pattern.

jwmatthews:
> @arapov, I understand the loop event might not be necessary, it's a good practice with a pipeline but absolutely not necessary, it all depends how we'd like to manage the addition of files to the workflow (the idea was to detect changes on the disk, submit a newly found file to the pipe and reiterate until times is up.).


There is no requirement for watching files on disk.  

The tool is intended to run once, it will look at an OCP 3 cluster and use the information present at that point in time.  It will perform translations to artifacts needed to bring what it can to the OCP 4 cluster.  The output of the tool will serve as a basis for a cluster administrator to begin tweaking their OCP 4 cluster.

I imagine a workflow of:
 - Ask user for information (details of their OCP 3 cluster, perhaps info on their OCP 4 cluster)
 - Collect all data required from OCP 3 (this will be fetching files from master/nodes perhaps as well as talking to OCP 3 API and obtaining configmaps/secrets/etc)
 - Iterate through specific OCP 4.x configuration items, translating data from OCP 3 to OCP 4

The big points I'd stress are to maintain good test coverage of the modules doing translation, as most of this tool is static in nature, it should lend itself well to testing.

gildub:
Such architecture is not needed, the reflection has helped walking through and offer different approaches to continue decouple the main actors.

Closing this one.

PR #79, opened 2019-04-29
export flag to ensure go mod installs deps

jmontleon:


jmontleon:
```

arapov:
The intention is good, though the implementation I don't like. We must not enforce developer's  environment for regular builds. This will be unfriendly.

This deserves a better build documentation and special build target in Makefile, so that a non-developer can actually get the binary.

arapov:
@alexander-demichev, let's discuss this tomorrow so that you can do the necessary changes to this pull request.

jmontleon:
@jwmatthews thank you clarifying.  It looks like all that's really needed is for GO111MODULE=on to be set for the build. `go mod vendor` is unnecessary.

PR #80, opened 2019-04-29
Add vendor

jmontleon:
Consider checking in vendor.

I'd be less interested in doing this on a project utilizing a multi-stage build for a container binary, but given this would be more likely to ship in an RPM than a container building dependencies individually would be unreasonably time consuming. Given you can't dep ensure/go mod vendor, etc. in brew you'll need to get dependencies in there one way or another.


arapov:
@jmontleon, Why would we do this? Why would anyone want to keep the entire copy of dependency sources?

That's exactly why go.mod and go.sum files and "go mod" created. I mean it's a clear NAK pull request.

jmontleon:
I mean I spelled it out in my original comment. If this is going to be shipped as an RPM you will either need to package all the dependencies so they are available in brew for build or vendor them.

arapov:
I'll need to refresh my knowledge about koji/brew then. I remember it was just good to do the builds from sources the way the spec file says. And unless we hit a real koji build show-stopper, I am against of keeping vendor/* in a tree.

eriknelson:
We have had this conversation on multiple projects and always come to the same conclusion, that vendor needs to be committed for the reasons Jason has laid out. The introduction of the new hot go dep management tool doesn't change any of the fundamental reasons why this is a problem.

arapov:
This merge is never too late to do then. Don't push it.

jmrodri:
@arapov I was beaten into submission on checking in vendor. I fought it tooth and nail and lost every time. Now I don't even bother. I just check in vendor. I agree with all of your comments and yet there is something nice about git pulling a go project and it JUST BUILDS! Often with using the dependency manager we would end up with different results (shouldn't but sometimes we did). With `dep` you can configure it to ONLY include the go code of your deps. By default it pulls in ALL of the code including non-go code and tests which really bloat your project.

https://github.com/openshift/ansible-service-broker/tree/master/vendor
https://github.com/operator-framework/operator-sdk/tree/master/vendor
https://github.com/openshift/service-catalog/tree/master/vendor
https://github.com/kubernetes/kubernetes/tree/master/vendor

It's a common pattern which is very counter intuitive but seems to solve the problem of making the project's build repeatable for everyone.

arapov:
I see your points, though we are fine to live non bloated live. At least until we need to get merry with Koji.

eriknelson:
I'm not happy about it either, but it's demonstrably the standard practice, and @jmontleon is suggesting this because he has been down this road more than once and he's speaking from painful experience.

jmrodri:
@arapov @alexander-demichev if you choose not to check in vendor, then what is the proposed solution for building in koji or brew? If the answer is hold off until you have to, that's fair. Otherwise, I'd be interested in seeing the solutions. Packaging all of the deps as separate packages isn't going to work well. I know @jmontleon tried that once and he's a freaking packaging wizard.

arapov:
Guys, *IF* we choose not to check in vendor, then we will have a solution for this and we will be happy to share it with everyone.

PR #82, opened 2019-04-30
OCP4 examples for OAuth and Secrets

gildub:


arapov:
That's good, though the only question I would have is - can we reuse this for testing? I know we have some of these in tests code already. 

If we can keep this examples in a proper place in a code tree and use them for tests - it would be awesome.

arapov:
I was afraid that this can be double effort to keep examples and keep same/similar examples we verify test results with. If it's not the case - approved.

Issue #76, opened 2019-04-29
Add ability to configure alternate configuration locations on OCP3 cluster


dymurray:
Currently this tool does not work with any `oc cluster up` environments. That isn't a huge deal since we don't support `oc cluster up` but for development it would be nice to test this against dev environments that use `oc cluster up`. 

Would be nice if https://github.com/fusor/cpma/blob/master/ocp3/ocp3.go#L104 was configurable in my `.cpma.yaml` file.

arapov:
We didn't pay much attention to oc cluster up. We will look into it.

@dymurray do you think the location of master and node config files can be different? Why would you want to change it?

dymurray:
Yeah I know they can be different :) When you run `oc cluster up` it creates a directory `openshfit.local.clusterup` which has `openshift-apiserver/master-config.yaml` and `node/node-config.yaml`. 

I hardcoded these paths in my fork:
```
// New instantiates and returns Config structure that represents OCP3 configuration
func New() *Config {
        return &Config{
                Masterf: "/tmp/openshift.local.clusterup/openshift-apiserver/master-config.yaml",
                Nodef:   "/tmp/openshift.local.clusterup/node/node-config.yaml",
        }
}
```

I think it makes sense to focus on `/etc/origin` since that is where production installs places these files. This is more just a nice to have thing for development.

PR #83, opened 2019-04-30
Add ability to use different master and node config paths

alexander-demichev:
Fixes https://github.com/fusor/cpma/issues/76

### DIFF
--- a/examples/cpma-config.example.yaml
+++ b/examples/cpma-config.example.yaml
@@ -1,7 +1,11 @@
 ---
-Source: "master-0.gildub1.lab.pnq2.cee.redhat.com" # can also point to directory
+Source: "master-0.example.com" # can also point to directory
 SSHCreds:
   Login: "root"
-  PrivateKey: "/home/gildub/.ssh/quicklab"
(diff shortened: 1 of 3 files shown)


examples/cpma-config.example.yaml
```diff
@@ -1,7 +1,10 @@
...
+Source: "master-0.example.com" # can also point to directory
 SSHCreds:
   Login: "root"
-  PrivateKey: "/home/gildub/.ssh/quicklab"
+  PrivateKey: "/home/example/.ssh/key"
   Port: 22
 OutputDir: "./data"
+# Masterf and Nodef are optional fields, use only if cluster was configured with different config locations
+Masterf: "/etc/origin/master/master-config.yaml"
+Nodef: "/etc/origin/node/node-config.yaml"
```
arapov: Patch looks good. One things before we merge:
- use more user friendly key name. E.g. "MasterConfigFile" is so much better than Masterf.

alexander-demichev: Done

PR #96, opened 2019-05-02
Rework fetch/decode

jmontleon:
Very WIP, trying to rework fetch/decode at the moment.

The current thinking is to:
1. Fetch the master config as the starting point.
1. Based on things we know and find from the master config start building a list of files to pull
1. Fetch files, read, and load data into a structure where we can access it conveniently
1. Transform that data into something OCP 4 can use
1. Write it out

One thing that's been noted:
@gildub mentioned that we might need to pull configs from multiple masters and nodes. I movedhostname off the Config(File) and onto the OCP3Cluster. If that's the case I'm not sure how we're to reconcile possible differences between masters and nodes.

Maybe for the cluster we have an array of hosts if need to be able to handle multiple and process the list of files per host and generate CR's per host within the cluster? I don't know. I would like to hear some cases for this and what would be the proper way to handle it if someone has some input.


jmontleon:
Note how this pulls the HTPasswd file, loads it into a field in a struct and writes it out to the secret later.

I believe this should be reusable for all types of files  (certs, keys, whatever) used in identity providers.
It also lets us Decode data from different types of config files (json, ini, whatever) and load it into a struct so we can write CR's.

gildub:
@jmontleon, for the record, thanks for you input, as discussed we're going to provide the decode flexibility.

PR #108, opened 2019-05-08
[WIP] Create Image CR for registry configuration

jmontleon:


alexander-demichev:
Overall looks looks good, are you planning to add tests later? Could you also update example config?

jmontleon:
Added WIP tag. Still working on it.

gildub:
@jmontleon, 

+1 for encapsulating with the registries.

Meanwhile I wonder why changing the decode core and using toml. 
The closest we stay from OCP3 libraries the better (unless there is a good reason to). We're already  mapping OCP3 and ideally @arapov  and I were thinking do the same with OCP structure and tap into their libraries.

gildub:
I sugest separating the registries code from the decode part in 2 different PRs.

gildub:
Sorry I posted and removed 2 comments that are not relevant.
This LGTM
+1 for tests

jmontleon:
Closing. This is included in #128.

Issue #107, opened 2019-05-08
Consider a classic ETL workflow


eriknelson:
Currently the code is structured as fetch, decode, transform, dump. I'm proposing we model the workflow to match the established [ETL pattern](https://en.wikipedia.org/wiki/Extract,_transform,_load). "Component" here refers to the abstraction described in #106:

* Extract: Fetch + decode the dependent config files for each component into usable data structures. Cache and reuse files that are shared between components (perhaps a `FileRetriever` type that checks for a cached copy, if miss, download).
* Validate: Ensure the data we've extracted looks correct to prevent garbage-in-garbage-out. This is important and not currently part of the workflow. @dymurray and I have already run into issues due to the lack of validate.
* Transform: Take the old data structures and turn them into the new structures
* Load (Dump, or Flush): Serialize the transformed structures into whatever destination it is that we desire, in this case, a set of CR files.

These methods should be defined on each component we write. A runner should accept a list of these components and execute each of them in sequence (could be parallelized via channels, but let's not pre-optimize).

gildub:
This is why I came up with the pipeline because it's a way of handling workflow like the ETL pattern.
From the beginning we have been working with a Fetch, Decode, Transform, and Dump approach in mind. I suppose it doesn't matter how many steps or operations there are. As long as we pass components along the chain.

The Validation operation is not in the pipe yet but will be. This has been on the back burner because besides validating YAML we need to find out how we can do more.

Yes we want to interface the components for the same operations, those being more FDTD than ETL but that's not an anti-pattern.

Now with more experience on the project we're more keen to deal with interfaces of interfaces.

eriknelson:
@gildub can you explain what you specifically mean by pipeline? The amount of steps definitely don't matter, but we think about the problem from a higher level, we know generally for all the relevant features in OCP4 we have to extract the old files, make sure they're okay, transform them to new ones, and write them out to files. We should strive to recognize the similarities with existing solved problems and leverage their solutions. In this case an ETL pattern is a well established approach for exactly what we're doing.

Re: the interfaces, we shouldn't use interfaces for the sake of it. It's useful because it let's you program to an interface of behavior, which means we can abstract this ETL concept, fill in the details for each component, and hand them all to a runner to execute in a uniform way.

The development model I want to get to is "oh I have a new feature I need to translate. Let me fill in X methods so I'm compliant with the transform interface, and then I can simply hand it to the runner and it will do that work".

[...]

Issue #106, opened 2019-05-08
Unit of translation should not be an individual config file


eriknelson:
In master I'm seeing a series of funcs (fetch, decode, translate, dump, or FDTD) run per remote config file (master, node). I'd like to propose an alternative model because this doesn't sufficiently address the problem of configuring OCP4 features, like container runtimes, which require multiple input configuration files.

The unit of translation should be on an OCP4 feature basis, not on a remote config basis. Let's call the OCP4 features components. For example, there is an `OAuth` component, an `SDN` component, and a `ContainerRuntime` component.

Ideally, you should be able to define the FDTD functions for each "component", and I want to fetch/decode the N config files that are relevant to that component.

gildub:
Putting OCP4 feature as the center of the component type is effectively good to handle N config files for 1 component.

jmrodri:
@eriknelson agreed. OCP4 seems to be what dictates how the code is structured. Doing it by component will yield the best outcome, it will make OCP3 a bit more complicated because you might have to fetch the same file a few times to get the appropriate info.

eriknelson:
@jmrodri, I expect a lot of components to share files. They should all probably ask some kind of FileManager type that either returns cached files or remkts files.

gildub:
@eriknelson, the io pkg provide a cache approach to sftp file retrieval.

gildub:
The units are now OCP4 components such as OAuth, SDN, etc.

PR #113, opened 2019-05-09
ETL Workflow with OCP V4 Components as interface

gildub:
OAuthConfig and DCNConfig are the new players.
NodeConfig (ConfigNode) is gone until it's broken down into future needed components.

https://github.com/fusor/cpma/issues/107
https://github.com/fusor/cpma/issues/106


cmd/root.go
```diff
@@ -48,20 +48,19 @@ var rootCmd = &cobra.Command{
...
-		ocpNode := new(ocp.ConfigNode)
-		ocpMaster.Add(source)
-		ocpNode.Add(source)
-		configs = append(configs, ocpMaster, ocpNode)
+		ocpOAuth := new(ocp.OAuthConfig)
+		ocpSDN := new(ocp.SDNConfig)
+		ocpOAuth.Add(source)
+		ocpSDN.Add(source)
+
+		configs = append(configs, ocpOAuth, ocpSDN)
```
jmrodri: root.go should not be creating these configs. I think you should have something like a `cpma.go` with a `Start` method that is launched from `root.go`. The `cpma.go` should be passed configurations based on the CLI parameters and start the process of doing migrations.

The `root.go` file should be relegated to handling CLI parameters then passing those to the `cpma.go`. Nothing more. Right now `root.go` has too much knowledge of the process.

gildub: I agree with that meanwhile others in the group don't.
So if you're against it please speak now or never!

gildub:
@jmrodri,

Are you sure we are talking about the same issues #.

Because the unit of translation is now an OCP4 component and not an individual configuration file.
The ETL pattern can now take place.

cmd/root.go
jmrodri: Yeah `root.go` should be parameters from CLI and then pass to a single entry that does all the work. I just like to separate the concerns as much as possible. Like I could theoretically rip out root.go and still be able to call the thing that does all the work from some other project, it's probably not a goal but it doesn't cost much to have a different file called from root.go. IMO

gildub: That's exactly what I thought.
The reason it that apparently it's the practice when using "github.com/spf13/cobra".

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
...
 // GetFile allows to mock file retrieval
 var GetFile = io.GetFile
 
-func (config *ConfigMaster) Add(hostname string) {
-	masterf := env.Config().GetString("MasterConfigFile")
+var source = env.Config().GetString("Source")
 
-	if masterf == "" {
-		masterf = "/etc/origin/master/master-config.yaml"
+func (sdnConfig *SDNConfig) Add(hostname string) {
```
eriknelson: It's not clear what the purpose of this method is, but from what I can see we're initializing input configs. IMO this should be done when the object is constructed.

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
...
+	SDN  sdn.NetworkCR
 }
 
 type Translator interface {
 	Add(string)
-	Decode()
-	Fetch(string)
+	Extract()
+	Transform()
 	GenYAML() ocp4.Manifests
```
eriknelson: This shouldn't be part of this interface. Semantically it doesn't make any sense on a `Translator` type, and the fact that it's `YAML` is an implementation detail.

Furthermore, it's a common practice with go to keep interfaces small and compose them. For example, see go's own [ReadWriter](https://golang.org/pkg/io/#ReadWriter).

There is a dividing line. It's not useful to take each of these methods and put them on individual interfaces. The atomic type here is the `Translator`. It's not useful to break it down any further.

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
...
-	config.ConfigFile.Hostname = hostname
-	config.ConfigFile.Path = nodef
+	oauthConfig.ConfigFile.Hostname = hostname
+	oauthConfig.ConfigFile.Path = path
 }
 
-func (config *ConfigMaster) Decode() {
-	config.OCP3.Decode(config.ConfigFile.Content)
+// Decode unmarshals OCP3 MasterConfig and sets OAuth
+func (oauthConfig *OAuthConfig) Decode() {
```
eriknelson: I don't think this requires its own method. It's part of `Extract`.

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
...
+	masterConfig := ocp3.MasterDecode(oauthConfig.Content)
+	oauthConfig.OCP3.OAuthConfig = masterConfig.OAuthConfig
 }
 
-func (config *ConfigNode) Decode() {
-	config.OCP3.Decode(config.ConfigFile.Content)
+func (sdnConfig *SDNConfig) Decode() {
```
eriknelson: Same as above.

pkg/ocp/ocp.go
```diff
@@ -84,48 +92,67 @@ func DumpManifests(outputDir string, manifests ocp4.Manifests) {
...
-	config.ConfigFile.Content = GetFile(config.Hostname, config.Path, localF)
-	logrus.Printf("File:Loaded: %s", localF)
-}
-
-func (config *ConfigNode) Fetch(outputDir string) {
-	localF := filepath.Join(outputDir, config.Hostname, config.Path)
-	config.ConfigFile.Content = GetFile(config.Hostname, config.Path, localF)
+func Fetch(configFile *ConfigFile) {
+	localF := filepath.Join(env.Config().GetString("OutputDir"), configFile.Hostname, configFile.Path)
+	configFile.Content = GetFile(configFile.Hostname, configFile.Path, localF)
```
eriknelson: +1 for `GetFile`. I'm picturing a lot of the components requiring similar files so the fact that we have some caching logic is a good thing.

pkg/ocp/ocp.go
```diff
@@ -84,48 +92,67 @@ func DumpManifests(outputDir string, manifests ocp4.Manifests) {
...
 	logrus.Printf("File:Loaded: %s", localF)
 }
 
-// GenYAML returns the list of translated CRDs
-func (config *ConfigMaster) GenYAML() ocp4.Manifests {
+// GenYAML returns the list of OAuth CRDs
+func (oauthConfig *OAuthConfig) GenYAML() ocp4.Manifests {
 	var manifests ocp4.Manifests
```
eriknelson: I'm concerned about a general type for `ocp4.Manifests`. Have we considered the fact that more than one ocp4 feature may need to provide their configuration on the **same CR**, which means they'd need to edit a shared output file. Perhaps we need to share a data structure between transforms that represents the total file output so that more than one `Translator` can mutate the same individual output file.

Example: FooFeature and BarFeature in OCP4 both are configured by writing to the same kube object, `FooBar`. How do we deal with that?

pkg/ocp4/oauth/openid.go
```diff
@@ -26,8 +26,8 @@ type identityProviderOpenID struct {
 	} `yaml:"openID"`
 }
 
-func buildOpenIDIP(serializer *json.Serializer, p configv1.IdentityProvider) (identityProviderOpenID, secrets.Secret) {
-	var idP identityProviderOpenID
+func buildOpenIDIP(serializer *json.Serializer, p configv1.IdentityProvider) (IdentityProviderOpenID, secrets.Secret) {
```
eriknelson: The casing of a function's argument is irrelevant to its accessibility, and as a matter of style, I think these should be lower cased.

pkg/ocp4/ocp4.go
```diff
@@ -36,78 +23,27 @@ const OCP4InstallMsg = `To install OCP4 run the installer as follow in order to
...
-	return manifests
-}
-
-func oauthManifests(oauthCRKind string, crd []byte, manifests []Manifest) []Manifest {
+func OAuthManifest(oauthCRKind string, crd []byte, manifests Manifests) Manifests {
 	if oauthCRKind != "" {
 		filename := manifestPrefix + "config-oauth.yaml"
-		manifest := Manifest{Name: filename, CRD: crd}
-		manifests = append(manifests, manifest)
+		m := Manifest{Name: filename, CRD: crd}
```
eriknelson: Please prefer to name you local variables with a meaningful name rather than `m`.

pkg/ocp4/ocp4.go
```diff
@@ -36,78 +23,27 @@ const OCP4InstallMsg = `To install OCP4 run the installer as follow in order to
...
 
-func oauthSecrets(ocp4MasterSecrets []secrets.Secret, manifests []Manifest) []Manifest {
-	for _, secret := range ocp4MasterSecrets {
-		filename := manifestPrefix + "config-secret-" + secret.Metadata.Name + ".yaml"
-		m := Manifest{Name: filename, CRD: secret.GenYAML()}
-		manifests = append(manifests, m)
-	}
-
+func SecretsManifest(secret secrets.Secret, crd []byte, manifests Manifests) Manifests {
+	filename := manifestPrefix + "config-secret-" + secret.Metadata.Name + ".yaml"
```
eriknelson: This is a pedantic comment, but `fmt.Sprintf` is a better alternative to adding strings to concatenate them.

gildub: Although not my style too. ;)

pkg/ocp4/oauth/openid.go
gildub: This can go in a different PR.

cmd/root.go
```diff
@@ -48,20 +48,19 @@ var rootCmd = &cobra.Command{
...
 		source := env.Config().GetString("Source")
-		outputDir := env.Config().GetString("OutputDir")
 
 		configs := []ocp.Translator{}
-		ocpMaster := new(ocp.ConfigMaster)
-		ocpNode := new(ocp.ConfigNode)
-		ocpMaster.Add(source)
-		ocpNode.Add(source)
-		configs = append(configs, ocpMaster, ocpNode)
+		ocpOAuth := new(ocp.OAuthConfig)
```
eriknelson: Just following up, this syntax isn't typically used in favor of literals, i.e. ocp.ConfigMaster{}. It's irrelevant here, but just passing that along.

cmd/root.go
jmrodri: > That's exactly what I thought.
> The reason it that apparently it's the practice when using "github.com/spf13/cobra".

I disagree, cobra does not dictate how you call the rest of your app. Let's be clear here, I'm not debating the need for the `root.go` file. And if cobra requires it to be called `root.go` that is also okay. What I'm against is this code living in `root.go`.

```golang
		ocpOAuth := new(ocp.OAuthConfig)
		ocpSDN := new(ocp.SDNConfig)
		ocpOAuth.Add(source)
		ocpSDN.Add(source)

		configs = append(configs, ocpOAuth, ocpSDN)

		for _, config := range configs {
			config.Extract()
			config.Transform()
			ocp.DumpManifests(config.GenYAML())
}
```

I'm of the opinion this should be in the core of your application. IMO `root.go` is NOT the core of your application. It is merely the "UI" for this application. Basically the struct that knows how to handle CLI parameters really doesn't care about what the core of the application needs. 

Basically I know to look in `cmd/` to see what parameters there are and how they are parsed. And I look to see how the application is started. 

So to satisfy the request I'm making it is as simple as moving the above code to another file in the `pkg/` directory, something like this:

```golang
MigrationAssistant struct {
 // the fields required to handle the transformations etc, I would defer to the POC that @eriknelson has been working on for this
}

func (m *MigrationAssistant) Start(params map[string]string) {
  // take in the parameters you need to start this application
}
```

Then in `root.go` I would do the following:

```golang
var rootCmd = &cobra.Command{
	Use:   "cpma",
	Short: "Helps migration cluster configuration of a OCP 3.x cluster to OCP 4.x",
	Long:  `Helps migration cluster configuration of a OCP 3.x cluster to OCP 4.x`,
	Run: func(cmd *cobra.Command, args []string) {
		env.InitConfig()
		env.InitLogger()

		// start the application
		ma := MigrationAssistant{}
		ma.Start(PASS_YOUR_PARAMS_HERE)
},
...
```

Do you see now? Basically `root.go` knows all about parsing parameters. It knows how to transform those CLI parameters into something that can be passed into the `Start` of the application. It's cleaner and easier to see where the stuff for cobra is and the stuff required by your application.

jmrodri:
> @jmrodri,
> 
> Are you sure we are talking about the same issues #.
> 
> Because the unit of translation is now an OCP4 component and not an individual configuration file.
> And the ETL pattern can now take place.
> 
> Please details if you disagree.

@gildub sure thing. Agreed the unit of translation is more correct, but the *structure* of the code is the crux of the issue. What @eriknelson and I are trying to do is organize the code such that the focus is on creating translators which are the bulk of the work. And so that it is clear to someone working in this project that they really should only need to add or update translators. The rest of the code is largely the infrastructure used to run translators.

eriknelson:
Want to help @gildub here so let's separate the blocking issues to merging this PR vs. longer term issues that should be filed:

Blockers:
* Be more specific about sources: https://github.com/fusor/cpma/pull/113#discussion_r282646447
* I don't think `Add` should be a method. Should consider what this is doing and maybe move the logic to `Extract` https://github.com/fusor/cpma/pull/113#discussion_r282647208
* These are not configs, they should be called `Translators` https://github.com/fusor/cpma/pull/113#discussion_r282647410
* Likewise: https://github.com/fusor/cpma/pull/113#discussion_r282647872
* `GenYAML` should not be part of the `Translator` interface: https://github.com/fusor/cpma/pull/113#discussion_r282667735
* `Decode` isn't doing much, its contents should be part of the `Extract` method: https://github.com/fusor/cpma/pull/113#discussion_r282674608, here as well: https://github.com/fusor/cpma/pull/113#discussion_r282674670

Long term issues:
* Reconsider what a `Translator` is called: https://github.com/fusor/cpma/pull/113#discussion_r282649141
* What is the purpose of `Add` on this type? https://github.com/fusor/cpma/pull/113#discussion_r282667033
* More than one component can output to the same CRs: https://github.com/fusor/cpma/pull/113#discussion_r282676836
* Reconsider top level app structure: https://github.com/fusor/cpma/pull/113#discussion_r282544936

cmd/root.go
gildub: I wanted to change that one and now I know why, thanks for the clarification.

This could go in the next PR.

gildub:
@jmrodri, I agree re the structure, to be done, I'll do a pass on that one ASAP.

cmd/root.go
```diff
@@ -48,20 +48,19 @@ var rootCmd = &cobra.Command{
...
+		ocpSDN := new(ocp.SDNConfig)
+		ocpOAuth.Add(source)
```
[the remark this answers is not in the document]
gildub: This is absolutely right, because of various undecided sources of endpoint including input for user (now deprecated, so far). 
This needs to be worked it.

cmd/root.go
```diff
@@ -48,20 +48,19 @@ var rootCmd = &cobra.Command{
...
+		ocpSDN.Add(source)
+
+		configs = append(configs, ocpOAuth, ocpSDN)
 
 		for _, config := range configs {
```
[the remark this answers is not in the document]
gildub: Done.

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
 	Content  []byte
 }
 
-type ConfigMaster struct {
+type OAuthConfig struct {
```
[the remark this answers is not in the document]
gildub: Done.

pkg/ocp/ocp.go
gildub: We are effectively not covering this use case yet.
But it might be needed so we need to anticipate. 
Will need to add it in following PR.

pkg/ocp/ocp.go
```diff
@@ -19,62 +23,66 @@ type ConfigFile struct {
...
-type ConfigNode struct {
+type SDNConfig struct {
 	ConfigFile
-	OCP3 ocp3.Node
-	OCP4 ocp4.Node
+	OCP3 configv1.MasterNetworkConfig
+	SDN  sdn.NetworkCR
 }
 
 type Translator interface {
```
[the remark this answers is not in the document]
gildub: Migration is better.
Another ugly one is CPMA !

eriknelson:
Long term issues linked above.

gildub:
[x]   These are not configs, they should be called Translators #113 (comment)
 [x]   Likewise: #113 (comment)
 [x]   GenYAML should not be part of the Translator interface: #113 (comment)

gildub:
[x] Decode isn't doing much, its contents should be part of the Extract method: #113 (comment), here as well: #113 (comment)

gildub:
We do need to clarify about sources but need some time and because the Add is involved I'd rather keep that one for following PR.

cmd/root.go
gildub: I think that's going to clarify it for good. Thanks

pkg/ocp4/oauth/openid.go
eriknelson: Please file an issue then and link here before we merge this.

eriknelson:
d

eriknelson:
Total mistake on my part closed this, apologies.

PR #112, opened 2019-05-09
Pass errors to top level

alexander-demichev:
This PR passes errors that have been caught to root.cmd
https://github.com/fusor/cpma/issues/109

PR #128, opened 2019-05-11
Use ETL style workflow rebased from master

jmontleon:


jwmatthews:
+1 to 'cpma report' and 'cpma transform' commands.

I like the structure changes and have an easier time following now with pkg/transform.

1. Do we want to parse several known files we think we will need each time in a common spot and then pass in?   Right now looks like we are reparsing master-config.yml in each transform component that needs it.  
   * Example:
     1. https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/oauth_transform.go#L27
     1. https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/sdn_transform.go#L70
2. Consider allowing Extract() to fetch all content it needs and returning, this way Run() is easier to test.  Assume something like ```Extract() map[string][interface{}]```
   * Below was example I saw, thinking pattern would benefit if Extract fetched everything needed and Transform was a pure pass me data and I will transform to what is needed., https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/oauth/htpasswd.go#L41
```golang
func buildHTPasswdIP(serializer *json.Serializer, p IdentityProvider) (IdentityProviderHTPasswd, secrets.Secret) {
	<snip>....
	src := filepath.Join(htpasswd.File)
	dst := filepath.Join(env.Config().GetString("OutputDir"), host, htpasswd.File)
	f := GetFile(host, src, dst)
}
 ```
3.  Smaller concern on structure of secrets package having specific details of things oauth would use. Realize isn't related to the ETL focus here, just more of a general question this feels strange to me.
https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/secrets/secrets.go#L11-L24
Having details of things 'oauth' knows about in 'secrets' package felt odd.   Is this desired?
```golang
package secrets

import (
	"github.com/sirupsen/logrus"
	"gopkg.in/yaml.v2"
)

// TODO: Comment exported functions and structures.
// We may want to unexport some...

type HTPasswdFileSecret struct {
	HTPasswd string `yaml:"htpasswd"`
}

type KeystoneFileSecret struct {
	Keystone string `yaml:"keystone"`
}

type LiteralSecret struct {
	ClientSecret string `yaml:"clientSecret"`
}
type BasicAuthFileSecret struct {
	BasicAuth string `yaml:"basicAuth"`
}
```
   
May be too early, but interesting to see how this pattern will lend itself as we add report capability in.  Looking for us to be able to capture most of the 'component' logic in one spot and be able to leverage that to generate a report as well as do the transform.   That may be a better exercise for later in week, just wanted to be sure we all are on same page that report will be important as well as transform.

eriknelson:
> Do we want to parse several known files we think we will need each time in a common spot and then pass in? Right now looks like we are reparsing master-config.yml in each transform component that needs it.

    Example:
        https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/oauth_transform.go#L27
        https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/sdn_transform.go#L70

I pictured some kind of a `FileManager` type that would be responsible for giving the transforms the file contents that they require, as called from the `Extract` method. Inside the type, it would implement the `get if we don't already have it, else go to the remote server and download the file, then return the contents`.

This might have an added benefit that the `FileManager` would be very easy to mock, so we could fake the file downloads for testing the `Extract` methods.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+)
+
+// OCP4InstallMsg message about using generated manifests
+const OCP4InstallMsg = `To install OCP4 run the installer as follow in order to add CRDs:
+' /openshift-install --dir $INSTALL_DIR create install-config'
+'./openshift-install --dir $INSTALL_DIR create manifests'
+# Copy generated CRD manifest files  to '$INSTALL_DIR/openshift/'
+# Edit them if needed, then run installation:
+'./openshift-install --dir $INSTALL_DIR  create cluster'`
+const MasterConfigFile = "/etc/origin/master/master-config.yaml"
```
eriknelson: Do we have any reason to believe these ocp3 file locations have ever changed? What if they changed between `3.x` versions? Normally I would say these should be passed by config values, but if they have never changed I think we can safely hardcode consts.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+
+type TransformRunner struct {
+	Config string
+}
+
+// GetFile allows to mock file retrieval
+var GetFile = io.GetFile
+
+func Start() {
+	config := Config{}
```
eriknelson: [ISSUE] I think we should put all of the configuration setup before running any of the `Transforms` behind a function, so that `Start` looks like:

```
config := loadConfig(/*possibly inputs*/)
//build and run transforms with config input
```

The end result of "config" will probably continue to become more complex, because we're going to keep adding new things to the config file as we learn more about the problem, and we also need to merge that with the cli provided options, and determine precedence if there is ever a conflict between either of those config sources (do we take the cli option, or the file option? Almost always, cli). Let's abstract that behind a function so we can change that without breaking main.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+
+type Config struct {
+	MasterConfigFile     string
+	NodeConfigFile       string
+	RegistriesConfigFile string
+	OutputDir            string
+	Hostname             string
+}
+
+type ManifestTransformOutput struct {
```
eriknelson: [BLOCKER] The concrete `TransformOutput` structs and methods should go into their own files, because it's likely we're going to add more of these. I would keep the general interface definitions here in this file, though.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+		RegistriesTransform{
+			Config: &config,
+		},
+	}); err != nil {
+		logrus.WithError(err).Fatalf("%s", err.Error())
+	}
+}
+
+// DumpManifests creates OCDs files
+func DumpManifests(manifests []Manifest) {
```
eriknelson: [BLOCKER] This is tightly coupled with the semantic purpose of a `ManifestTransformOutput`, so I would move this method into the same file you put that struct and its interface implementations.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+		os.MkdirAll(path.Dir(maniftestfile), 0755)
+		err := ioutil.WriteFile(maniftestfile, manifest.CRD, 0644)
+		logrus.Printf("CRD:Added: %s", maniftestfile)
+		if err != nil {
+			logrus.Panic(err)
+		}
+	}
+}
+
+func (config *Config) Fetch(path string) []byte {
```
eriknelson: [ISSUE] This doesn't feel like it belongs in this file. I've talked about this before, but it would be really nice to have some kind of a type called a `FileManager` who's role it is to deliver requested file contents to the extract methods. In the simplest case, this could just be a vanilla function that has a signature of `GetFile(fileName string)`, but I suspect it will continue to get more complex then that, especially when you consider the fact that you may have to ask for a file from a specific host in the multihost case. Maybe that's just another argument to `GetFile`, and pure functions are easier to test, but this is also a function that, by its nature, has externalities to be considered (other hosts). Just kind of dumping my thoughts, I would not move this for now.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	dst := filepath.Join(config.OutputDir, config.Hostname, path)
+	f := GetFile(config.Hostname, path, dst)
+	logrus.Printf("File:Loaded: %s", dst)
+
+	return f
+}
+
+type Extraction interface {
```
eriknelson: :tada:

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	Transform() (TransformOutput, error)
+	Validate() error
+}
+
+type Transform interface {
```
eriknelson: I originally had called this a transform, but I've been uncertain of the correct language here that sufficiently describes the concept. @jwmatthews you've called this a `Component` in a lot of our conversations. What I'd like for this interface to represent is one of the encapsulated features to be moved over from ocp3 to ocp4. I settled on calling this a `Migration` on the most recent version of my poc, which represents the flow of actions that happen for a given feature. Part of that is the ETL, but we already identified another possible step we want to include that is "inspect the extraction and determine if the customer has made changes, and apply a <green>, <yellow>, or <red> status" to the incoming data. I don't have a full grasp of how this work work in my mind, but I think this is the natural place to possible put that step.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	Extract() Extraction
+}
+
+type TransformOutput interface {
```
eriknelson: I really like having this, it gives us optionality for how we want to output our ocp4, in-memory data structures, and we already had a request for a new output before this even got added. +1

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	Flush() error
+}
+
+func (m ManifestTransformOutput) Flush() error {
```
eriknelson: Should go in another file with the rest of the `ManifestTransformOutput`. I would not say a new pkg, it makes sense to have them as part of the transform pkg, but new file.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	DumpManifests(m.Manifests)
+	return nil
+}
+
+func (r TransformRunner) Transform(transforms []Transform) error {
+	logrus.Info("TransformRunner::Transform")
+
+	// For each transform, extract the data, validate it, and run the transform.
+	// Handle any errors, and finally flush the output to it's desired destination
+	// NOTE: This should be parallelized with channels unless the transforms have
```
eriknelson: Because of the org @jmontleon has here, we've now consolidated a **single** place to implement parallelization via go channels to these ~20 lines. This is a huge win IMO.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	// some dependency on the outputs of others
+	for _, transform := range transforms {
```
eriknelson: If we want to add another required method in the transform series, we now have one place to call it, and it's just a matter of implementing it on each of the transforms. +1.

pkg/transform/transform.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	return nil
+}
+
+func NewTransformRunner(config Config) *TransformRunner {
+	return &TransformRunner{}
+}
+
+func HandleError(err error) error {
```
eriknelson: [ISSUE] I had this in the POC as the top level error handler just to illustrate typical go error handling and bubbling. We should reconsider this (I know @alexander-demichev has a PR out that is seeking to address it).

pkg/transform/registries_transform.go
```diff
@@ -0,0 +1,86 @@
...
+
+type Registries struct {
+	List []string `toml:"registries"`
+}
+
+type ImageCR struct {
+	APIVersion string   `yaml:"apiVersion"`
+	Kind       string   `yaml:"kind"`
+	Metadata   Metadata `yaml:"metadata"`
+	Spec       struct {
```
eriknelson: [BLOCKER] Shoudln't inline this type. It's a pain in the ass, but I think they should be broken out into their own types.

pkg/transform/registries_transform.go
```diff
@@ -0,0 +1,86 @@
...
+	content := c.Config.Fetch(c.Config.RegistriesConfigFile)
+	var extraction RegistriesExtraction
+	if _, err := toml.Decode(string(content), &extraction); err != nil {
+		HandleError(err)
+	}
+	return extraction
+}
+
+func (c RegistriesExtraction) Validate() error {
+	return nil // Simulate fine
```
eriknelson: [ISSUE] We need to implement these and test the validations, stat.

pkg/transform/registries_transform.go
```diff
@@ -0,0 +1,86 @@
...
+	manifests = append(manifests, manifest)
+
+	return ManifestTransformOutput{
+		Manifests: manifests,
+	}, nil
+}
+
+func (c RegistriesTransform) Extract() Extraction {
+	logrus.Info("RegistriesTransform::Extract")
+	content := c.Config.Fetch(c.Config.RegistriesConfigFile)
```
eriknelson: [ISSUE] These calls are where I'd like to see that `FileManager` type I mentioned elsewhere. I will file this to try to describe my thoughts.

pkg/transform/registries_transform.go
```diff
@@ -0,0 +1,86 @@
...
+	var extraction RegistriesExtraction
```
eriknelson: I think this is a big win. Extract the data and decode it into an `Extraction`. Let the `Extraction` define what it means to `Transform` and `Validate` itself. +1

pkg/transform/transform.go
eriknelson: On top of that, you could inject a `MockFileManager` for testing extract functions with dummy data, that complies with the `FileManager` interface.

pkg/transform/transform.go
jmrodri: Not knowing the full scope, `Transform` seems okay for now. If it changes to `Migration` that's probably okay too.

pkg/transform/registries_transform.go
jmrodri: Own type but it's okay in the same file.

eriknelson:
@jwmatthews I think @jmontleon addressed your second point since your comment:

> Consider allowing Extract() to fetch all content it needs and returning, this way Run() is easier to test. Assume something like Extract() map[string][interface{}]

We now have strong `Extraction` types. These are a better alternative to the `map[string]interface{}` that I was admittedly throwing around (and you pointed out to me (:man_facepalming:). I think his code here fixes that.

> Having details of things 'oauth' knows about in 'secrets' package felt odd. Is this desired?

+1, I think if they're types that only make sense within the oauth context, they should probably move into the oauth pkg.

> Looking for us to be able to capture most of the 'component' logic in one spot and be able to leverage that to generate a report as well as do the transform.

IMO, this PR gets us very close to this. Imagine 2 different `TransformOutput` types. The runner can tell the transform which one to use on this line: https://github.com/fusor/cpma/pull/128/files#diff-cd30b5e0e7a0599257ceca4b46512c7cR138, based on a cli siwtch.

```
if(config.ReportMode) {
  `output, err := extraction.Transform(TransformOutputType.Report)
} else {
  `output, err := extraction.Transform(TransformOutputType.CRFiles)
}
```

Transform would return the correct outputter and it'd get flushed.

pkg/transform/registries_transform.go
eriknelson: [BLOCKER] I would actually add a logrus warning to all the transforms with a log message saying "VALIDATION NOT IMPLEMENTED" or something to that effect. We want that to annoy us every time we run this until it gets implemented.

eriknelson:
[ISSUE] We shouldn't have more than one cobra file based on if you want a "report" or "CR output files". This should be an enum set on the config, and we should write a new `TransformOutput` depending on the value the value of the config. Alluded to it here: https://github.com/fusor/cpma/pull/128#issuecomment-491563595

pkg/transform/transform.go
jwmatthews: I'm not aware of location differences with `openshift-ansible`.  We have seen differences between `openshift-ansible` and `oc cluster up` for file paths.  

To be safe, I think there is value in allowing all of these locations to be specified in config file and overridden.  

As @eriknelson mentioned with blockers and non-blockers, I would consider this a non-blocker and something we could improve in a follow up issue.

jwmatthews:
I like the changes with 
```golang
type Extraction interface {
	Transform() (TransformOutput, error)
	Validate() error
}

type Transform interface {
	Extract() Extraction
}
```

Question, @eriknelson do you have thoughts/example of what it would look like to write a unit test for one of the transform components.  For example, say we want to write a test for `pkg/transform/registries_transform_test.go`.

I'm a little unsure of how testing ```Transform()``` would work out.  
We'd receive an instance of ```TransformOutput``` and could invoke ```Flush()```, but not sure what the thoughts are to use that in a test.

I think it'd be worth it to capture at least 1 simple unit test on a _transform.go in this PR just to ensure we are in good shape for building out those tests in future PR.  The pieces I am most interested in are the workflow of what do we pass into the _transform component to test it, how to we get the data it produced and validate it.  

Ok if the test itself is simplistic as long as it captures the workflow/pattern so others can see the intent and help to improve testing in subsequent PRs.

jmontleon:
@eriknelson regarding the multiple cobra files I'm not very familiar with cobra and reading their documentation it seemed to be what it was suggesting.

Particularly here:

jmontleon:
@eriknelson regarding the multiple cobra files I'm not very familiar with cobra and reading their documentation it seemed to be what it was suggesting.

Particularly here:
https://github.com/spf13/cobra#getting-started
https://github.com/spf13/cobra#create-additional-commands

In the first it looks like they're suggesting multiple files. In the last it kind of looks like the same.

pkg/transform/registries_transform.go
jmontleon: Removed this here, sdn_transform.go, and pkg/oauth/oauth.go. I think I got them all.

pkg/transform/registries_transform.go
jmontleon: Added logrus warnings to the 3 validation methods that they are not implemented.

pkg/transform/transform.go
jmontleon: Agree. I was hoping we could reuse @alexander-demichev's work. Unsure though if it's going to be a major rework. I can try to help with that if necessary.

pkg/transform/transform.go
jmontleon: Would you put this in a separate pkg, or just a different file in the same pkg - something like transform_outputs.go?

pkg/transform/transform.go
jmontleon: Added a LoadConfig to do this and I think it addresses the concern. If not let me know what more should be done here!

pkg/transform/transform.go
alexander-demichev: @jmontleon Feel free to do anything you think is necessary. I will rebase or rework PR with error handling.

eriknelson:
> I'm a little unsure of how testing Transform() would work out. We'd receive an instance of TransformOutput and could invoke Flush(), but not sure what the thoughts are to use that in a test. I think we could use some kind of 

@jwmatthews That's is a good question that I hadn't really considered and should have thought more about. One option that comes to mind would be to make the `Flush` method for the `TransportOutput` implementations a default pkg level function. That would allow us to override it for test purposes and flush to something in memory that we'd then be able to compare against expected values. I take a stab at an example based off this branch.

pkg/transform/transform.go
eriknelson: [ISSUE]

pkg/transform/transform.go
eriknelson: If we're going to use a pkg level default function for flush so that we can override it, you'd only be able to do that if you kept them in the same package. I think it makes sense for them all to be part of a transform pkg.

eriknelson:
Consolidating feedback:

**Blockers**
* Transport output implementations should be in their own files: https://github.com/fusor/cpma/pull/128#discussion_r283117826, `DumpManifests` should also belong in the same file: https://github.com/fusor/cpma/pull/128#discussion_r283117839, Flush implementation should go in the same spot, although I think we're going to make this a default pkg function for test purposes. Should stay in the transform package: https://github.com/fusor/cpma/pull/128#discussion_r283118052
* ~~Shouldn't use inline types in struct definitions~~
* ~~Warn validate methods are not implemented~~

**Issues to be filed**
* Remote config file paths should be configurable via cli/cpma configfile overrides: https://github.com/fusor/cpma/pull/128#discussion_r283117690
* Fetch shouldn't be part of transform, might want some kind of a `FileManager`: https://github.com/fusor/cpma/pull/128#discussion_r283117899, https://github.com/fusor/cpma/pull/128#discussion_r283118222
* Reconsider error handling, @alexander-demichev is working on that with his PR here: #112 
* Investigate best practices around multiple cobra files: https://github.com/fusor/cpma/pull/128#issuecomment-491564445, https://github.com/fusor/cpma/pull/128#issuecomment-491595033
* Need to rework `Flush` on the transport outputs to be package level default fns so that they can be overridden under test https://github.com/fusor/cpma/pull/128#issuecomment-491611443

pkg/transform/transform.go
jmontleon: I moved ManifestTransformOutput stuff to its own file. I think this is resolved.

pkg/transform/transform.go
jmontleon: I moved ManifestTransformOutput stuff to its own file. I think this is resolved.

eriknelson:
@jwmatthews @jmontleon

Played a bit with unit testing these and implemented what I described before. Moved the contents of flush to an overridable package level, private function:

https://github.com/eriknelson/cpma/commit/bdd651a0616350d178fd39274bd4b80b3abf0424#diff-cd30b5e0e7a0599257ceca4b46512c7cR126

I'm having trouble creating basic input data with `loadTestIdentityProviders` (it's blowing up somewhere in the parsers), and I'm also just not familiar with what the expected output should be, but this should illustrate the basic structure of how you would test one of these transforms:

https://github.com/eriknelson/cpma/blob/jmontleon-rebased-rework/pkg/transform/oauth_transform_test.go#L53

Key part is where we override the flush method to hand back the resulting manifests over a channel to the main test routine rather than writing the `Manifest` type to a CR file. Kick off the extraction in the background (https://github.com/eriknelson/cpma/blob/jmontleon-rebased-rework/pkg/transform/oauth_transform_test.go#L70) and wait until we get back the resulting manifests (https://github.com/eriknelson/cpma/blob/jmontleon-rebased-rework/pkg/transform/oauth_transform_test.go#L78), at which time we should compare the actual manifests we got back from the output with expected manifests (https://github.com/eriknelson/cpma/blob/jmontleon-rebased-rework/pkg/transform/oauth_transform_test.go#L81).

eriknelson:
@jmontleon fixed all my blockers afaict. @jwmatthews do you think we should block this merge until we get a working example test, or should we merge this and file an issue to add the example test?

I'm okay with merging this + issue because there have been no regressions to existing tests.

jwmatthews:
> @jmontleon fixed all my blockers afaict. @jwmatthews do you think we should block this merge until we get a working example test, or should we merge this and file an issue to add the example test?
> 
> I'm okay with merging this + issue because there have been no regressions to existing tests, and there's a clear path forward with unit testing this architecture based on my above example.

@eriknelson defer to you on decision of when to merge.

My ask is to work with @jmontleon and get a single unit test on a transform written soon.  Ok if that happens early in week after this PR is merged. 

Please also include in issues to write the refactor of “oauth” details in secrets package, https://github.com/jmontleon/cpma/blob/rebased-rework/pkg/transform/secrets/secrets.go#L11-L24

eriknelson:
@jwmatthews ack re: the unit test, we should be able to follow up with that tomorrow. We just need to work through some parsing issues with input data unrelated to the core organization of how to test one of these. I don't consider it a risk and I just didn't have time to dig into it.

Added your issue re: oauth details living in the secrets package.

pkg/transform/oauth/github_test.go
```diff
@@ -1,20 +1,48 @@
...
 )
 
 func TestTransformMasterConfigGithub(t *testing.T) {
 	file := "testdata/github-test-master-config.yaml"
 	content, _ := ioutil.ReadFile(file)
-	masterV3 := ocp3.MasterDecode(content)
+
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	var masterV3 configv1.MasterConfig
+	_, _, _ = serializer.Decode(content, nil, &masterV3)
```
dymurray: Are we eating the error here? Any reason to do that?

pkg/transform/oauth/keystone_test.go
```diff
@@ -17,7 +21,30 @@ func TestTransformMasterConfigKeystone(t *testing.T) {
 
 	file := "testdata/keystone-test-master-config.yaml"
 	content, _ := ioutil.ReadFile(file)
-	masterV3 := ocp3.MasterDecode(content)
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	var masterV3 configv1.MasterConfig
+	_, _, _ = serializer.Decode(content, nil, &masterV3)
```
dymurray: Looks like this is done in every unit test for the most part. So if it's intentional ignore.

pkg/transform/oauth/github_test.go
alexander-demichev: @dymurray I will take care of error handling in tests here https://github.com/fusor/cpma/pull/112 after https://github.com/fusor/cpma/pull/128 gets merged.

pkg/transform/oauth/oauth.go
```diff
@@ -107,9 +123,11 @@ func Transform(oauthconfig *configv1.OAuthConfig) (*OAuthCRD, []secrets.Secret,
 			}
 		default:
 			logrus.Infof("Can't handle %s OAuth kind", kind)
+			continue
 		}
 		oauthCrd.Spec.IdentityProviders = append(oauthCrd.Spec.IdentityProviders, idP)
-		if secret != (secrets.Secret{}) {
+
+		if secret.Metadata.Name != "htpasswd_auth-secret" || p.Kind == "HTPasswdPasswordIdentityProvider" {
```
dymurray: Consider setting `htpasswd_auth-secret` and `HTPasswdPasswordIdentityProvider` as constants

pkg/transform/oauth_transform.go
```diff
@@ -0,0 +1,93 @@
...
+	var htContent []byte
+
+	_, _, err := serializer.Decode(content, nil, &masterConfig)
+	if err != nil {
+		HandleError(err)
+	}
+
+	if masterConfig.OAuthConfig != nil {
+		for _, identityProvider := range masterConfig.OAuthConfig.IdentityProviders {
+			providerJSON, _ := identityProvider.Provider.MarshalJSON()
```
dymurray: Don't eat `MarshalJSON` error

pkg/transform/transform.go
dymurray: Just worth noting, the current implementation allows you to set the location of the master-config: https://github.com/fusor/cpma/blob/master/pkg/migration/migration.go#L77

pkg/transform/oauth/github_test.go
dymurray: Awesome, thanks!

pkg/transform/oauth/keystone_test.go
jmontleon: It is intentional. There's probably some work to be done to clean this up. 

MasterDecode and Extract don't exactly map 1 to 1. This gets the tests working, but I don't think it's ideal.

pkg/transform/oauth/oauth.go
jmontleon: Hopefully this looks better now.

pkg/transform/transform.go
```diff
@@ -0,0 +1,139 @@
...
+	if err := transformRunner.Transform([]Transform{
+		OAuthTransform{
+			Config: &config,
+		},
+		SDNTransform{
+			Config: &config,
+		},
+		RegistriesTransform{
+			Config: &config,
+		},
```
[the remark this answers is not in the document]
jmontleon: I'm not sure if you're suggesting the Config shouldn't hold the file contents? It doesn't...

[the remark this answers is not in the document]
eriknelson: We might be conflating two different concepts that could be described as `Config`:

* The program takes user input for how it should run between the cli and a cpma configuration file. We should read the cli inputs and the config files contents, merge the two onto a data structure that represents `UserInput`, and then that should be made available to each of the transforms so they can customize their behavior based on the user's input.
* Remote cluster config, like the master and node configs.

The first absolutely should be made available to the transforms for the reasons outlined above. I think you might be thinking of the second case. You'd definitely be right that the remote cluster data shouldn't be passed into the transforms from the outside; they should be extracting it themselves in the `Extract` method and putting that decoded data onto the specific `Extraction` type. That's the specific purpose for that method & type.

This block is correct from my standpoint.

PR #129, opened 2019-05-13
Add registries validation

jmontleon:


jmontleon:
Spaces are fixed. I used a patch to reapply my changes. I always seem to get spaces when I do that and have to resave files. Sorry about that.

PR #136, opened 2019-05-14
Fix lint errors

jmontleon:


jmontleon:
@eriknelson are you ok with the function renames? Lint was complaining that Transform.TransformRunner was a stutter and suggested changing it to Transform.Runner, etc.

eriknelson:
@jmontleon I was often guilty of lint problems for this very reason with the broker. It still feels unnatural to me but I understand and agree with the reasoning.

jmontleon:
Think it should be fixed now.

eriknelson:
Fixed for me.

PR #138, opened 2019-05-14
Fix Makefile lint

jmontleon:
Using golang-1.12.2-1.fc30.x86_64 the existing lint command does not work.
I' would be curious to know if anyone using an older version finds this does or doesn't work for them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,9 @@ test: ## Test the project
 	GO111MODULE=on go test ./...
 
 lint: ## Run golint
-	@golint -set_exit_status $(addsuffix /... , $(SOURCE_DIRS))
+	@for i in $(SOURCE_DIRS); \
+		do golint -set_exit_status $(addprefix ./, $(addsuffix /... , $$i)); \
+		done
 
 fmt: ## Run go fmt
 	@gofmt -d $(SOURCES)


jmrodri:
Unfortunately, you can't use the packaged `golint` it is way too old. Fedora 29 uses commit hash https://github.com/golang/lint/commit/c7bacac which is from 2016.

Fedora 30 uses the same hash:
```
# rpm -q golint
golint-0-0.10.gitc7bacac.fc30.x86_64
```

Those versions of golint don't support multiple directories on the CLI. You need to install golint using go get.

```
go get -u golang.org/x/lint/golint
```

This version supports `cmd/... pkg/...` as parameters.

jmontleon:
Thanks for the explanation @jmrodri 

@gildub I think @eriknelson was seeing a lint error I wasn't and that I don't think even Code Climate saw. Now I am wondering if this was also due to the differing golint versions. 

It looks like no one has paid much mind to the RPM almost since it was initially created, other than to fix a build issue and disable tests almost a year ago. https://src.fedoraproject.org/rpms/golint/commits/master

I'm normally in favor of RPM's, but this one could do with an update. I wonder if there is a technical challenge or it just needs a new maintainer.

jmontleon:
Closing this for now. We can reopen or submit a new PR if for some reason we decide we want to support really old golint versions.

Right now I'd almost prefer checking for the rpm and suggesting uninstalling it (although even if it's installed depending how you have your path set up it doesn't mean it will be used.) golint does not have a version command unless I'm just blind so I don't see an easy way to determine the version by running golint -v or something along those lines.

PR #142, opened 2019-05-14
mv env to pkg, allow overriding config file locations

jmontleon:


eriknelson:
Checked with @alexander-demichev in slack to confirm that there wasn't a specific, unknown reason why the env pkg was at the top level of the project. We moved to `pkg` to conform with go convention, this makes sense to me. ACK

PR #148, opened 2019-05-14
Add oauth, sdn, and registries transform unit test

### LINKED ISSUE #135: The new transforms need an example unit test
Spawned from https://github.com/fusor/cpma/pull/128#issuecomment-491611443

I provided an example for how to accomplish this here: https://github.com/eriknelson/cpma/commit/bdd651a0616350d178fd39274bd4b80b3abf0424

Specifically this is how I would convert `Flush` into overridable behavior so it becomes testable: https://github.com/eriknelson/cpma/commit/bdd651a0616350d178fd39274bd4b80b3abf0424

I'm not 100% satisfied with pkg level default functions; it doesn't feel like that's a sustainable model for dependency injection (DI). Something to consider, I was chatting with another kubernetes go developer about how people deal with DI in go and there have been some promising proposals in major projects that maybe we could leverage? https://github.com/kubernetes-sigs/controller-runtime/pull/285

Pragmatically I do not think we should adopt this, but it's an interesting option to explore, should we choose to do so.

jmontleon:
Closes #135

### DIFF
--- a/pkg/transform/oauth_transform_test.go
+++ b/pkg/transform/oauth_transform_test.go
@@ -1,0 +1,323 @@
+package transform
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"io/ioutil"
+	"testing"
+
+	"github.com/fusor/cpma/pkg/transform/oauth"
+	"github.com/fusor/cpma/pkg/transform/secrets"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	configv1 "github.com/openshift/api/legacyconfig/v1"
+	k8sjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func loadTestIdentityProviders() []oauth.IdentityProvider {
+	// TODO: Something is broken here in a way that it's causing the translaters
+	// to fail. Need some help with creating test identiy providers in a way
+	// that won't crash the translator
+
+	// Build example identity providers, this is straight copy pasted from
+	// oauth test, IMO this loading of example identity providers should be
+	// some shared test helper
+	file := "testdata/bulk-test-master-config.yaml" // File copied into transform pkg testdata
+	content, _ := ioutil.ReadFile(file)
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	var masterV3 configv1.MasterConfig
+	_, _, _ = serializer.Decode(content, nil, &masterV3)
+
+	var identityProviders []oauth.IdentityProvider
+	for _, identityProvider := range masterV3.OAuthConfig.IdentityProviders {
+		providerJSON, _ := identityProvider.Provider.MarshalJSON()
+		provider := oauth.Provider{}
+		json.Unmarshal(providerJSON, &provider)
+
+		identityProviders = append(identityProviders,
+			oauth.IdentityProvider{
+				Kind:            provider.Kind,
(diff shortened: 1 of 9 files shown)


jmontleon:
TODO's and t.Log(f)'s removed.

eriknelson:
@jmontleon re: my remark about equality, I took a closer look at this and I think it's actually doing what we expect it to do. The objects we're testing (`Manifest`) are simple https://github.com/fusor/cpma/pull/148/files#diff-074e096a46e8203beecc17975713e0a9R259, they're just a `Name` (`string`) and `CRD` (`[]byte`). They are not nested types, and don't contain any pointers.

Additionally, I moved one of the expected values into a different slot in the array, and that failed the test. That tells me the arrays are ordered with the same objects at each index. < That concern goes away once we split out the array into individual tests. The transform output would return an array of one object, that we'd compare to the single identity provider that we passed as input.

I think this is doing everything we expect it to do, so I'm ready to ack it with an issue that we need to return to this and write a Test function for each identity provider rather than having them all in one test function.

pkg/transform/oauth_transform_test.go
```diff
@@ -0,0 +1,305 @@
...
+	content, _ := ioutil.ReadFile(file)
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	var masterV3 configv1.MasterConfig
+	_, _, _ = serializer.Decode(content, nil, &masterV3)
+
+	var identityProviders []oauth.IdentityProvider
+	for _, identityProvider := range masterV3.OAuthConfig.IdentityProviders {
+		providerJSON, _ := identityProvider.Provider.MarshalJSON()
+		provider := oauth.Provider{}
+		json.Unmarshal(providerJSON, &provider)
```
jwmatthews: Is this pattern normal for parsing the master-config.yml.
Seemed odd to:
 - Parse YAML to Struct
 - Extract a piece of Struct and marshal to JSON
 - Unmarshal JSON to a Struct

Realize it's a copy/paste from an existing block of test code and not related to testing the ETL pattern.
More just a general question, is there a reason for doing this?

jmontleon: I'll have to defer to @eriknelson about the 'normacly' of it.

jmrodri: After digging into this I think this is the best way. But we really should put a comment above it to describe WHY we're doing something that looks so weird :)

`identityProvider.Provider` is of type `runtime.RawExtension`. As you can see here https://github.com/openshift/api/blob/master/legacyconfig/v1/types.go#L852

`runtime.RawExtension` has 2 attributes:

```golang
	Raw []byte `protobuf:"bytes,1,opt,name=raw"`
	Object Object `json:"-"`
```
https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/types.go#L94-L102

So in order to get the `Kind`, etc. needed to fill in the `oauth.IdentityProvider`, that `[]byte` needs to be converted into something useful.

That's where the MarshalJSON comes into play.

Unless there exists utility methods in `apimachinery` to do this for us, looks like this seems to be the best way. Also it's a test :)

gildub: That's exactly why. 
Earlier in the project, we also talked about being able to tap into `apimachinery` to get full support of the `kind` genre.
  
With more coming components to cover it's something it might worth looking into it to save us time redefining the target for each new component wrapped inside each RawExtension. 

Any pointer to reach out for one of the OCP/Component (like Auth) Team's member to do a knowledge transfer to fast forward use of their API?

eriknelson:
I think we've got sufficient consensus on this to merge.

PR #152, opened 2019-05-15
Fix empty secret for basic auth

alexander-demichev:
BasicAuth provider always creates an empty secret, this PR should fix.
The way how we handle empty secrets now maybe be improved, condition on line 148 can get bigger with time.
I believe we can accept this PR as a temporary fix and focus on more important translation logic.

### DIFF
--- a/pkg/transform/oauth/oauth.go
+++ b/pkg/transform/oauth/oauth.go
@@ -89,6 +89,8 @@ const (
 	httpdAuthSecret = "htpasswd_auth-secret"
 	// httpProviderKind is the htpasswd idp type string
 	httpProviderKind = "HTPasswdPasswordIdentityProvider"
+	// basicAuthProviderKind is the basic auth idp type string
+	basicAuthProviderKind = "BasicAuthPasswordIdentityProvider"
 )
 
 // Translate converts OCPv3 OAuth to OCPv4 OAuth Custom Resources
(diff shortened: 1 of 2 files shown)


pkg/transform/oauth/oauth.go
```diff
@@ -143,7 +145,7 @@ func Translate(identityProviders []IdentityProvider) (*CRD, []secrets.Secret, er
 		}
 		oauthCrd.Spec.IdentityProviders = append(oauthCrd.Spec.IdentityProviders, idP)
 
-		if secret.Metadata.Name != httpdAuthSecret || p.Kind == httpProviderKind {
+		if (secret.Metadata.Name != httpdAuthSecret || p.Kind == httpProviderKind) && p.Kind != basicAuthProviderKind {
```
gildub: I might have missed something but I would use the above switch entries to handle those cases.

pkg/transform/oauth/oauth_test.go
```diff
@@ -106,6 +106,6 @@ func TestGenYAML(t *testing.T) {
 	CRD := crd.GenYAML()
 	expectedYaml, _ := ioutil.ReadFile("testdata/expected-bulk-test-masterconfig-oauth.yaml")
 
-	assert.Equal(t, 10, len(manifests))
+	assert.Equal(t, 9, len(manifests))
```
gildub: Sorry but I can't see why the number has decreased?

alexander-demichev: One of secrets here was empty

jmontleon: Thanks for fixing this. I couldn't understand why the empty secret was being generated. I definitely saw it yesterday when I was working on the oauth transform test.

pkg/transform/oauth/oauth.go
jmontleon: Yes. this needs some attention, I think. I added the the line above with the intention of coming back to figure out how to better handle this.

jmontleon:
I'm OK to ACK this, while also acknowledging we need to come work on it, or we can try to improve it now. Either is fine with me.

alexander-demichev:
@jmontleon sure, no problem here

PR #153, opened 2019-05-15
Add lint to CI

alexander-demichev:


Makefile
```diff
@@ -27,6 +27,8 @@ ci: build test
 test: ## Test the project
 	GO111MODULE=on go test ./...
 
+ci-lint: lint fmt vet
+
```
jmrodri: Any reason not to add this to the ci target? 

```
ci: lint fmt vet build test
```

Devs can just run them individually locally, most often `make lint` is run.

.travis.yml
```diff
@@ -18,5 +18,10 @@ matrix:
         directories:
           - $HOME/Library/Caches/go-build
           - $HOME/gopath/pkg/mod
+
+before_install:
+  - env GO111MODULE=off go get -u golang.org/x/lint/golint
+
 script:
+  - make ci-lint
```
jmrodri: +1 to adding this to travis, see my other comment about possibly combining the 2 targets.

eriknelson: Are you upvoting splitting the linting out from the build/test? Or just the general idea we should add this stuff to travis?

alexander-demichev:
@eriknelson PR is ready for another review

Issue #155, opened 2019-05-15
Create configmaps for oauth providers


alexander-demichev:
Few oAuth providers require config maps that hold a the certificate authority, we need to implement their creation.

This should be handled here https://github.com/fusor/cpma/blob/master/pkg/transform/oauth/ldap.go#L48 and the logic for creating config maps will be similar to secrets

PR #156, opened 2019-05-15
Get and Load Basic Auth Cert and Key

jmontleon:


jmontleon:
@alexander-demichev does this look (in)sane to you?

I think we can get these all knocked out like this.

pkg/transform/oauth/github_test.go
```diff
@@ -39,17 +39,18 @@ func TestTransformMasterConfigGithub(t *testing.T) {
...
-				nil,
-				identityProvider.UseAsChallenger,
-				identityProvider.UseAsLogin,
+				Kind:            provider.Kind,
+				APIVersion:      provider.APIVersion,
+				MappingMethod:   identityProvider.MappingMethod,
+				Name:            identityProvider.Name,
+				Provider:        identityProvider.Provider,
+				HTFileName:      provider.File,
+				HTFileData:      nil,
```
eriknelson: If these are pointer types, they will naturally default to nil, so you can safely just leave them out of the initializer.

jmontleon: True. Now that it's not in the format it was these can go away. I'll remove them.

jmontleon:
This may not be quite right. Please don't merge it. Going from the issue Alex filed (and I only saw after writing this PR) it may not be secrets, but rather configmaps for CA's.

https://github.com/fusor/cpma/issues/155

jmontleon:
If the CA part is wrong I'll rip it out and get the rest in for downloading the files, and loading the Cert, Key to their respective secrets at least.

jmontleon:
CA portion writing a secret for it needs to be removed. CA's need to be stored as a configmap.
https://docs.openshift.com/container-platform/4.1/authentication/identity_providers/configuring-basic-authentication-identity-provider.html#identity-provider-basic-authentication-CR_configuring-basic-authentication-identity-provider

jmontleon:
Keystone was fixed from the fetch rework for keystone and htpasswd so we could unblock tests.

It looks like they're the only two using certs and keys. 

The remainder use CA's which will take some additional work to create a config-map instead of a secret, but the file is downloaded and content loaded into the struct with this PR so it's ready to be transformed.

PR #159, opened 2019-05-16
OAuth: Providers report when no value

gildub:
Need to filter cases such as when Keystone has no values for CA/Cert if fails the whole OAuth Extract operation. 

Filtering those cases using, for now, the existing set of warnings to report them.

This also adds `fileExists` to io, for more robustness.


gildub:
This has been covered with
https://github.com/fusor/cpma/pull/156

Issue #154, opened 2019-05-15
Factorize Master, Node, configuration file unmarshalling.


gildub:
I thought someone mentioned it but I cannot find in PR#128 a comment about it.

Anyway, since lots of components are sub set of the same configuration file, shouldn't we rely on a helper to handle each time we tap into a MasterConfig, or NodeConfig, etc.

For instance, these 2 blocks have the same use and both share configV1.MasterConfig: 
https://github.com/fusor/cpma/blob/master/pkg/transform/oauth_transform.go#L57-L66
https://github.com/fusor/cpma/blob/master/pkg/transform/sdn_transform.go#L103-L110

Helpers could be something like:
https://github.com/fusor/cpma/blob/8c7523c9ec9a90959fae074a9ab02b6533f7f835/pkg/ocp3/ocp3.go#L24-L49
 
Your thoughts?

jmontleon:
I think the more we can cut down on duplicate code the better. 

As to where it would live, maybe we could get all our decoders for various formats in pkg/util/decode.go or something. I'll defer to @eriknelson as to where might be best.

Even better if we can make it more generic to handle any json content. Likewise, TOML, YAML, INI, and anything else we're going to have to deal with. All the better if we can use it to eliminate all the places I copy/pasted it into tests too. Was just commenting to the same effect on @alexander-demichev's error handling PR as well.

eriknelson:
> Anyway, since lots of components are sub set of the same configuration file, shouldn't we rely on a helper to handle each time we tap into a MasterConfig, or NodeConfig, etc.

> I think the more we can cut down on duplicate code the better.

Absolutely. I could almost picture a `config` pkg with all the data structures that you would decode into, as well as the decoder logic (that could even be in a `decode` pkg found at `config/decode/`). Some pseudocode usage of this would look something like:

**oauth_transform.go**
```
import (
  "github.com/fusor/cpma/pkg/config"
  "github.com/fusor/cpma/pkg/config/decode"
  ...
)

func Extract() {
    var masterConfig config.MasterConfig
    decode.MasterConfigDecode(masterConfig, GetFile(env.Config.MasterConfigFile))
    // Build an Extraction based on the data we've now got in the masterConfig
}
```

Issue #133, opened 2019-05-13
Consider file retrieval as an independent feature


eriknelson:
Spawned from PR #128

We have a generic requirement that most (all?) transforms are going to need to get remote file contents as part of their `Extract` methods, to be decoded and stored on an `Extraction` object. It makes sense that we have a dedicated part of the codebase who's only responsibility is to provide this feature. Additionally, we should be able to easily mock it with something that provides `testdata` sourced contents so we can test `Extract` methods well.

If the above is true, we should move `Fetch` [out of transform/transform.go](https://github.com/fusor/cpma/blob/master/pkg/transform/transform.go#L97), and into its own file, and I would advocate a different pkg than `transform`. This is the kind of thing that typically ends up in a `util` package if it's a handful of functions, but if it grows to be more than that, I think it makes sense in its own `file` pkg (name TBD).

It should handle file caching since most transforms will be requesting the same files from it, I know we do that today, but just speccing that here. Additionally, @gildub has pointed out that it's likely that we will need the ability to provide the host as an argument, since it's possible this may need to support a multi-host scenario.

Finally, it could possibly be useful to represent this concept with something like a `FileManager` struct rather than a handful of functions. I don't have a reason for this right now, but I've certainly seen it done for something like this in the past.

At the very least, `Fetch` as it is today should be moved out of that transform file and into something more focused on the concept of file retrieval, and it shouldn't be called from anywhere but transform `Extract` methods, since that's the only place that should be downloading remote files.

Some Q's to be answered:
* Is it possible that in a multi-master scenario, the master config files would differ between masters? If that's not possible, just pick one, and we don't have to care about specifying the host.
* Do we need to get files from nodes as well? If that's true, we know we'll need to at least get files from a master and a node, so that's more than one host, and we'd have to accept the host as an argument here as well. It seems valuable to do this regardless of the answer to the Q's.

gildub:
https://github.com/fusor/cpma/issues/278 is partly replacing this issue as things have evolved.

The Fetch part has already been decoupled from the Transform, and going even furtherto separate remote file fetching as a complete separate pkg, the user being able to fetch files independently.

That said the FileManager idea is still valid and would happen within the latest context.

PR #160, opened 2019-05-16
Config own package

gildub:
Closes https://github.com/fusor/cpma/issues/154

Helps with https://github.com/fusor/cpma/issues/133  by getting Fetch() outside of transform.go.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,0 +1,44 @@
+package config
+
+import (
+	"path/filepath"
+
+	"github.com/fusor/cpma/pkg/env"
+	"github.com/fusor/cpma/pkg/io"
+	"github.com/sirupsen/logrus"
+)
+
+// reference:
+// https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html
+
+// TODO: we may want to be OCP3 minor version aware here
+
+// Config contains CPMA configuration information
+type Config struct {
+	OutputDir string
+	Hostname  string
+}
+
+// Fetch files from the OCP3 cluster
+func (c *Config) Fetch(path string) ([]byte, error) {
+	dst := filepath.Join(c.OutputDir, c.Hostname, path)
+	logrus.Infof("Fetching file: %s", dst)
+	f, err := io.GetFile(c.Hostname, path, dst)
+	if err != nil {
+		return nil, err
+	}
+	logrus.Infof("File:loaded: %v", dst)
+
+	return f, nil
+}
+
+// LoadConfig collects and stores configuration for CPMA
+func LoadConfig() Config {
+	logrus.Info("Loaded config")
+
+	config := Config{}
+	config.OutputDir = env.Config().GetString("OutputDir")
+	config.Hostname = env.Config().GetString("Source")
+
+	return config
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -34,9 +34,8 @@ func (c *Config) Fetch(path string) ([]byte, error) {
 func LoadConfig() Config {
 	logrus.Info("Loaded config")
 
-	config := Config{}
-	config.OutputDir = env.Config().GetString("OutputDir")
-	config.Hostname = env.Config().GetString("Source")
-
-	return config
+	return Config{
+		OutputDir: env.Config().GetString("OutputDir"),
+		Hostname:  env.Config().GetString("Source"),
+	}
 }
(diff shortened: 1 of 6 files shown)


pkg/transform/oauth_transform.go
```diff
@@ -70,16 +69,15 @@ func (e OAuthTransform) Extract() (Extraction, error) {
...
 	}
 
-	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
-	var masterConfig configv1.MasterConfig
-	var extraction OAuthExtraction
-	var htContent, caContent, crtContent, keyContent []byte
-
-	_, _, err = serializer.Decode(content, nil, &masterConfig)
+	var masterConfig *configv1.MasterConfig
+	masterConfig, err = decode.MasterConfig(content)
```
jmrodri: Instead of using var I would simply do:
```golang
masterConfig, err := decode.MasterConfig(content)
```

pkg/transform/oauth_transform.go
```diff
@@ -70,16 +69,15 @@ func (e OAuthTransform) Extract() (Extraction, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
+	var extraction OAuthExtraction
+	var htContent, caContent, crtContent, keyContent []byte
```
jmrodri: If you aren't going to use these outside the loop I would simply declare them using the `:=`, see next comment.

jmrodri: On line 95 changed it to be `htContent, err := e.Config.Fetch(provider.File)`. Typically when you create a var outside a block you intend to use that variable elsewhere.

pkg/config/config.go
```diff
@@ -0,0 +1,42 @@
...
+	return f, nil
+}
+
+// LoadConfig collects and stores configuration for CPMA
+func LoadConfig() Config {
+	logrus.Info("Loaded config")
+
+	config := Config{}
+	config.OutputDir = env.Config().GetString("OutputDir")
+	config.Hostname = env.Config().GetString("Source")
```
jmrodri: I prefer specifying the fields during the `{}`:
```golang
config := Config{
    OutputDir: env.Config().GetString("OutputDir")
    Hostname: env.Config().GetString("Source")
}
```

Actually I'd probably go a step further and replace the return with:

```golang
return  Config{
    OutputDir: env.Config().GetString("OutputDir")
    Hostname: env.Config().GetString("Source")
}
```

That might be going to far for some but at a minimum I'd use the first form of creating the instance.

pkg/config/decode/decode.go
```diff
@@ -0,0 +1,38 @@
...
+	configv1.InstallLegacy(scheme.Scheme)
+}
+
+// MasterConfig unmarshals OCP3 Master
+// There is no known differences between OCP3 minor versions of the master config
+func MasterConfig(content []byte) (*configv1.MasterConfig, error) {
+	var masterConfig = new(configv1.MasterConfig)
+
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	_, _, err := serializer.Decode(content, nil, masterConfig)
```
jmrodri: Interesting seeing the use of `new()`. I've always done

```golang
masterConfig := configv1.MasterConfig{}
...
serializer.Decode(content, nil, &masterConfig)
```
That is clearly a style preference, no need to change, just interesting to see your style.

pkg/transform/oauth_transform.go
gildub: Yes, I agree, just that I didn't pass on all existing assignments.

pkg/transform/oauth_transform.go
gildub: Done

pkg/config/decode/decode.go
gildub: In this case I don't really see it as a different style approach.
Because when a pointer is needed `new` provides one to use along.
No need to remember this is not a pointer so one must use the address.

pkg/config/config.go
gildub: I appreciate the style difference.
And I'm happy to use the fields during the {}.
As long as we're consistent across the project.

That said I don't like using it for readability reason, especially when there are embedded struct(s).
I'm extrapolating, the good practice of defining structs separately instead of embedded sub structs, and apply it to assignments of as well.
And so, no matter if an object has or not sub structs, in order to be consistent,  use struct method for assignments.

gildub: I suppose depending on the context one form might better than the other.
Anyway I've changed it to be consistent and also because with the return it's more concise.

eriknelson:
Spoke with @jmrodri who confirmed this is good to merge.

PR #161, opened 2019-05-16
Refactor tests

### LINKED ISSUE #111: Consider using the Golang structured test paradigm
https://github.com/automationbroker/bundle-lib/blob/master/registries/adapters/dockerhub_adapter_test.go#L35-L147

Basically create an array of test cases to cover edge cases and easy cases as well.  The actual struct can be tailored to your individual test needs. The important part is the array.

```
	testCases := []struct {
		name        string
		c           Configuration
		expected    []string
		expectederr bool
		handlerFunc http.HandlerFunc
}{
...
}
```

Then create your test loop:

```
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			// get test server
			serv := httptest.NewServer(tc.handlerFunc)
			defer serv.Close()

			// use the test server's url
			dockerHubLoginURL = strings.Join([]string{serv.URL, "/v2/users/login/"}, "")
			dockerHubRepoImages = strings.Join([]string{serv.URL,
				"/v2/repositories/%v/?page_size=100"}, "")
			dockerHubManifestURL = strings.Join([]string{serv.URL, "/v2/%v/manifests/%v"}, "")

			// create the adapter we want to test
			dha := DockerHubAdapter{Config: tc.c}

			// test the GetImageNames method
			output, err := dha.GetImageNames()

			if tc.expectederr {
				if !assert.Error(t, err) {
					t.Fatal(err)
				}
				assert.NotEmpty(t, err.Error())
			} else if err != nil {
				t.Fatalf("unexpected error during test: %v\n", err)
			}

			errmsg := fmt.Sprintf("%s returned the wrong value", tc.name)
			assert.Equal(t, tc.expected, output, errmsg)
		})
}
```

alexander-demichev:
This PR should make our tests more structured.

Closes #111

### DIFF
--- a/pkg/env/env_test.go
+++ b/pkg/env/env_test.go
@@ -16,32 +16,35 @@ func TestInitConfig(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	expectedHomeDir, _ := homedir.Dir()
-	actualHomeDir := viperConfig.GetString("home")
-	assert.Equal(t, expectedHomeDir, actualHomeDir)
+	expectedHomeDir, err := homedir.Dir()
+	if err != nil {
+		t.Fatal(err)
+	}
 
-	expectedConfigFilePath := ConfigFile
-	actualConfigFilePath := viperConfig.ConfigFileUsed()
-	assert.Equal(t, expectedConfigFilePath, actualConfigFilePath)
+	testCases := []struct {
+		name                   string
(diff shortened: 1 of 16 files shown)


jmrodri:
@alexander-demichev AWESOME JOB! Kudos! :tada:

PR #164, opened 2019-05-17
Make etl common for transform and report

alexander-demichev:
This PR introduces usage of report package. I think report should follow ETL, because we are extracting configuration and make a report based on it, so I decided that both transform and report could share same logic.

SDN is partially Day 1 operation. Report takes SDN config and splits it into 2 parts: a CR for migrating network plugin type and yaml snippet for install config that contains IP adresses. Report also provides a `readme.md`.


gildub:
I understand the idea of having 2 workflows and therefore share between them.
Meanwhile I see lots of duplication here.

So do we really need a complete different workflow for Report?
Or is it just a special case of Transform?
Because the same Extract part is needed, the Load part is different as we don't have to produce CRDs manifests but we have to report any differences discovered.

The first part of the report requirements is about the control plane, so that is potentially covered during the transform pass where we already discover any differences (obsolete/deprecated value, etc). Those warnings can be propagated to feed the report message.

For the 2nd part, the application workload would require it's own parsing. Could we have 2 different types of transform?

eriknelson:
To add report support, which is just another type out output. We should only have to write another TransformOutput implementation that the Transform method builds depending on if we're in report or transform mode as ready by the config. That's the entire point of the TransformOutput interface.

I haven't had a chance to dig into this yet and probably won't until Monday, but I think this PR needs some work. Some very rough Transform pseudocode:

```
func (e Extraction) Transform() {
switch config.OutputMode {
OutputMode.Report: // enum
  return e.buildReportOutput()
OutputMode.Manifest:
  return e.buildManifestOutput()
}
```

jmrodri:
@eriknelson @gildub seems to me this PR should be closed and a new one be opened that adds more of the TransformOutput. Report shouldn't be a whole new path IMO at least from the top. It should be the same work until the very end when the report is generated.

eriknelson:
> Report shouldn't be a whole new path IMO at least from the top

+1, this fundamentally circumvents the point of the ETL architecture we worked to establish.

alexander-demichev:
@eriknelson @jmrodri There is no problem in closing this PR and opening a new one. I'm glad we started discussion about report. Maybe we should talk about report implementation on Monday/Tuesday, I would really like to see some initial part of report merged asap, so it doesn't block us from adding support for new config parts.

gildub:
Do we agree (or not, then discuss more) that a report is just another output format?

In such case case one must first indicate the format output expected at the end of the transform chain. 
It also means each component is going to be analyzed in regards of the planning needs (control plane, application workloads), which would involve different parsing phases.

jmrodri:
Closing this PR based on the above review comments.

eriknelson:
I think we're in agreement this should be closed until we have a design doc that illustrates how we plan to handle reports and exactly what they should contain. @jmontleon is currently working on that.

PR #166, opened 2019-05-17
Remove Config

jmontleon:


jmontleon:
env/viper and Config seem redundant to me.

gildub:
@jmontleon , @alexander-demichev , keep in mind that using a global source (hostname) is only working for us at the moment but we are more likely to have a dynamic where each config file has a non global hostname as source and therefore we'll use a Config (or maybe FileConfig or so, to distinguish if from env) for each.

jmrodri:
This one seems weird. While it does seem like Config is being duplicated and repeated I think it does serve a purpose. I think the idea here is give the Transforms whatever they need to operate at creation. This PR seems to make things more "global" which isn't the direction I would expect. It will make things more brittle IMO.

gildub:
We definitely need to agree on the definition of a Config.

A Config is the starting point of specific to each transformation, it holds the hostname, configuration file tuple. 

The pitfall is that initially we associated a cluster entry point ("endpoint") to the Config and use it as a general case. But let's add another node (the node-config) and see how that would not work as we'd need 2 Configs.

PR #170, opened 2019-05-20
Create config map for oauth providers CA

alexander-demichev:
I will add tests and fix CI in next commit

Closes https://github.com/fusor/cpma/issues/155

### DIFF
--- a/pkg/transform/configmaps/configmaps.go
+++ b/pkg/transform/configmaps/configmaps.go
@@ -1,0 +1,57 @@
+package configmaps
+
+import (
+	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
+)
+
+// ConfigMap represent configmap definition
+type ConfigMap struct {
+	APIVersion string   `yaml:"apiVersion"`
+	Kind       string   `yaml:"kind"`
+	Metadata   MetaData `yaml:"metadata"`
+	Data       Data     `yaml:"data"`
+}
+
+// MetaData configmap's metadata
+type MetaData struct {
+	Name      string `yaml:"name"`
+	Namespace string `yaml:"namespace"`
+}
+
+// Data contains CA
+type Data struct {
+	CAData string `yaml:"ca"`
+}
+
+const (
+	// APIVersion is the apiVersion string
+	APIVersion = "v1"
+	// Kind is config map resource type
+	Kind = "ConfigMap"
+)
+
+// GenConfigMap generates a secret
+func GenConfigMap(name string, namespace string, CAData []byte) *ConfigMap {
+	return &ConfigMap{
+		APIVersion: APIVersion,
+		Data: Data{
+			CAData: string(CAData),
+		},
+		Kind: Kind,
+		Metadata: MetaData{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+}
+
+// GenYAML returns a YAML of the configMap
+func (configMap *ConfigMap) GenYAML() ([]byte, error) {
+	yamlBytes, err := yaml.Marshal(&configMap)
+	if err != nil {
+		logrus.Debugf("Error in config map, config map - %+v", configMap)
+		return nil, err
+	}
+	return yamlBytes, nil
+}
--- a/pkg/transform/oauth/basicauth.go
+++ b/pkg/transform/oauth/basicauth.go
@@ -3,6 +3,7 @@ package oauth
 import (
 	"encoding/base64"
 
+	"github.com/fusor/cpma/pkg/transform/configmaps"
 	"github.com/fusor/cpma/pkg/transform/secrets"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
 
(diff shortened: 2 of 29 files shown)


pkg/transform/oauth_transform.go
```diff
@@ -55,6 +56,17 @@ func (e OAuthExtraction) Transform() (Output, error) {
 			m := Manifest{Name: filename, CRD: secretCR}
 			manifests = append(manifests, m)
 		}
+
```
alexander-demichev: I'm thinking of a way to avoid repeting L59-69 in golang, they are similiar to L49-58

pkg/transform/configmaps/configmaps.go
```diff
@@ -0,0 +1,59 @@
...
+	// APIVersion is the apiVersion string
+	APIVersion = "v1"
+	// Namespace is the config map's default namespace string
+	Namespace = "openshift-config"
+	// Kind is config map resource type
+	Kind = "ConfigMap"
+)
+
+// GenConfigMap generates a secret
+func GenConfigMap(name string, CAData []byte) *ConfigMap {
```
gildub: Maybe we need to move to a more generic function and pass the Namespace as parameters to the builder so we could create ConfigMaps for any namespace?

alexander-demichev: Makes sense, I will change it

pkg/transform/oauth_transform.go
gildub: Not sure if it worth creating a Manifest builder.

pkg/transform/oauth/basicauth.go
```diff
@@ -44,12 +45,17 @@ func buildBasicAuthIP(serializer *json.Serializer, p IdentityProvider) (Identity
 	idP.MappingMethod = p.MappingMethod
 	idP.BasicAuth.URL = basicAuth.URL
 
+	if basicAuth.CA != "" {
```
eriknelson: How often will there not be a CA value here?

There's going to have to be some logic somewhere that is calling these buildXIP functions and doing things with the return values. That also means there's logic somewhere that's going to have to check that `caConfigMap` map value to see if it's actually empty or not, because we're giving back a zero value'd object.

It would be much easier to represent the absence of a `caConfigMap` as `nil` rather than an empty one. That way, the code that's interested in this value can do a simply nil check rather than having to inspect the type to determine if it's empty or not.

pkg/transform/oauth/basicauth.go
```diff
@@ -44,12 +45,17 @@ func buildBasicAuthIP(serializer *json.Serializer, p IdentityProvider) (Identity
...
+		caConfigmap = configmaps.GenConfigMap("basicauth-configmap", "openshift-config", p.CAData)
+		idP.BasicAuth.CA.Name = caConfigmap.Metadata.Name
+	}
+
 	certSecretName := p.Name + "-client-cert-secret"
 	idP.BasicAuth.TLSClientCert.Name = certSecretName
 	encoded := base64.StdEncoding.EncodeToString(p.CrtData)
 	certSecret, err = secrets.GenSecret(certSecretName, encoded, "openshift-config", "basicauth")
```
eriknelson: I'm noticing the `openshift-config` namespace is hardcoded all over the place. Is it possible this value may differ between clusters? I.E., some clusters may have these secrets and certs in a different location? Regardless, this should be a const instead of a magic string.

eriknelson: We can file an issue for this if we don't already have one, since it wasn't this PR that introduced that.

pkg/transform/oauth/oauth.go
```diff
@@ -144,10 +148,15 @@ func Translate(identityProviders []IdentityProvider) (*CRD, []secrets.Secret, er
 			continue
 		}
 
+		// Check if config map is not empty
+		if caConfigMap != (configmaps.ConfigMap{}) {
```
eriknelson: I would be careful with this kind of statement. This probably does what we're asking it to do here, but testing for "Equality" on two types can be much more complex than it appears to be on the surface (deep vs. shallow checks).

I think the better golike/kubelike way of representing the absence of a ConfigMap is to have them return nil, and not an empty struct. This pattern is used all over go, as an example, with returning errors from functions. It's also done all over kubernetes to represent an optional type, used to distinguish the absence of something vs. the presence of something that exists, but is empty.

pkg/transform/oauth_transform.go
eriknelson: If you see this crop up again, that may be time to introduce something like a generic interface:

```
type ManifestGenerator interface {
  GenYAML() string
}
```

And have one function handle this. Rule of 3 is useful, if you see code like this show up 3 times, it's a pretty good indication it's time to refactor to something more generic because there's obviously shared behavior in there somewhere.

pkg/transform/oauth/basicauth.go
alexander-demichev: "openshift-config" namespace is required for all configmaps and secrets related to oAuth https://docs.openshift.com/container-platform/4.1/authentication/identity_providers/configuring-ldap-identity-provider.html

eriknelson: ack. It's still good practice to define this string value with a const and use that value as opposed to raw strings.

Issue #169, opened 2019-05-20
Omit optional values if they are empty


alexander-demichev:
We should omit optional values if they are empty. For example:
```
identityProviders:
  - name: my_remote_basic_auth_provider 
    challenge: true 
    login: true 
    mappingMethod: claim 
    provider:
      apiVersion: v1
      kind: BasicAuthPasswordIdentityProvider
      url: https://www.example.com/remote-idp 
      ca: /path/to/ca.file # OPTIONAL
      certFile: /path/to/client.crt  # OPTIONAL
      keyFile: /path/to/client.key # OPTIONAL
```
certFile, ca and keyFile can be empty and we should not generate secrets and have them in translated CR.

alexander-demichev:
https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#BasicAuthPasswordIdentityProvider

PR #171, opened 2019-05-21
 Omit empty values for oauth provider

alexander-demichev:
This PR is based on https://github.com/fusor/cpma/pull/170, I'll rebase after it gets merged.

Closes https://github.com/fusor/cpma/issues/169

### DIFF
--- a/pkg/transform/oauth/basicauth.go
+++ b/pkg/transform/oauth/basicauth.go
@@ -17,20 +17,22 @@ type IdentityProviderBasicAuth struct {
 }
 
 // BasicAuth provider specific data
+// BasicAuth provider specific data
 type BasicAuth struct {
-	URL           string        `yaml:"url"`
-	CA            CA            `yaml:"ca"`
-	TLSClientCert TLSClientCert `yaml:"tlsClientCert"`
-	TLSClientKey  TLSClientKey  `yaml:"tlsClientKey"`
+	URL           string         `yaml:"url"`
+	CA            *CA            `yaml:"ca,omitempty"`
+	TLSClientCert *TLSClientCert `yaml:"tlsClientCert,omitempty"`
+	TLSClientKey  *TLSClientKey  `yaml:"tlsClientKey,omitempty"`
 }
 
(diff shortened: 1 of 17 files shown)


pkg/transform/oauth/basicauth.go
```diff
@@ -16,21 +16,23 @@ type IdentityProviderBasicAuth struct {
...
 	var (
-		err                   error
-		idP                   IdentityProviderBasicAuth
-		certSecret, keySecret *secrets.Secret
-		caConfigmap           *configmaps.ConfigMap
-		basicAuth             configv1.BasicAuthPasswordIdentityProvider
+		err         error
+		idP         IdentityProviderBasicAuth
+		certSecret  = &secrets.Secret{}
+		keySecret   = &secrets.Secret{}
```
alexander-demichev: We should return nil instead of empty secret, I will change it in next PR

pkg/transform/oauth/basicauth.go
```diff
@@ -16,21 +16,23 @@ type IdentityProviderBasicAuth struct {
...
 
+// BasicAuth provider specific data
 // BasicAuth provider specific data
 type BasicAuth struct {
-	URL           string        `yaml:"url"`
-	CA            CA            `yaml:"ca"`
-	TLSClientCert TLSClientCert `yaml:"tlsClientCert"`
-	TLSClientKey  TLSClientKey  `yaml:"tlsClientKey"`
+	URL           string         `yaml:"url"`
+	CA            *CA            `yaml:"ca,omitempty"`
```
jmontleon: Are we safe from nil pointer dereference's etc. with these. I'm fine with it as long as we're not as risk of stuff unexpectedly blowing up.

alexander-demichev: I added tests where this values are empty, should be safe. Is there any other way to omit empty values when working with structures?

jmontleon: If there are tests to cover it I'm good. Thanks!

PR #174, opened 2019-05-22
Validate oAuth providers

alexander-demichev:


pkg/transform/oauth/github.go
```diff
@@ -66,3 +67,32 @@ func buildGitHubIP(serializer *json.Serializer, p IdentityProvider) (IdentityPro
 
 	return idP, *secret, caConfigmap, nil
 }
+
+// ValidateGithubProvider validate github identity provider
+func ValidateGithubProvider(serializer *json.Serializer, p IdentityProvider) error {
```
jmrodri: Having specific validation methods for each specific provider seems reasonable. It will definitely be easy to maintain.

pkg/transform/oauth_transform.go
```diff
@@ -150,8 +152,22 @@ func (e OAuthTransform) Extract() (Extraction, error) {
...
-	logrus.Warn("Oauth Transform Validation Not Implmeneted")
-	return nil // Simulate fine
+	var err error
+
+	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	for _, identityProvider := range e.IdentityProviders {
+		switch identityProvider.Kind {
+		case "GitHubIdentityProvider":
+			err = oauth.ValidateGithubProvider(serializer, identityProvider)
```
jmrodri: I don't think `OAuthExtraction` should know about the specific Provider types. I feel like it should simply call:
```golang
for _, identityProvider := range e.IdentityProviders {
    oauth.Validate(serializer, identityProvider)
}
```

Then `Validate` should live along side `Translate` in `oauth.go`. 

https://github.com/fusor/cpma/blob/master/pkg/transform/oauth/oauth.go#L89-L164

This way `Validate` can look similar to `Translate` by knowing to look at the Kind and decide which way to validate the individual providers.

eriknelson: I don't feel too strongly about this either way, but if we're going to have the extraction essentially delegate `Transform` to the `oauth` package, it makes sense to do the same for `Validate`.

alexander-demichev: @eriknelson @jmrodri I initially placed `Validate(oauthProviders)` in `ouath.go` when started writing this, but then moved it to `ouath_transform.go` because I wasn't sure where it should be. Now with your feedback, I know `ouath.go` is better, so I'm moving it back:)

pkg/transform/oauth/oauth.go
```diff
@@ -171,3 +173,57 @@ func (oauth *CRD) GenYAML() ([]byte, error) {
...
+		case "LDAPPasswordIdentityProvider":
+			err = validateLDAPProvider(serializer, identityProvider)
+		case "OpenIDIdentityProvider":
+			err = validateOpenIDProvider(serializer, identityProvider)
+		case "RequestHeaderIdentityProvider":
+			err = validateRequestHeaderProvider(serializer, identityProvider)
+		}
+		if err != nil {
+			return err
+		}
```
jmrodri: Do you want to stop processing if there is an error in ANY of the providers?  If so, okay. If not, the for loop definitely will exit the first time one of the providers fails to validate the rest will be skipped.

PR #175, opened 2019-05-22
Changed type of secretType parametr to enum.

### LINKED ISSUE #158: secretType with GenSecret and the switch in buildData should be an enum
https://github.com/fusor/cpma/blob/master/pkg/transform/secrets/secrets.go#L68

It's preferable to use an enum to represent something like this so it uses a strong type and won't accept a `secretType` that isn't part of the possible set (the compiler will stop it in its tracks).

Go doesn't have enums, but there are idiomatic ways they can be approximated:

```
type Direction int

const (
    North Direction = iota
    East
    South
    West
)

func (d Direction) String() string {
    return [...]string{"North", "East", "South", "West"}[d]
}
```

https://yourbasic.org/golang/iota/

Danil-Grigorev:
Fixed tests after changes, updated usage of the `GenSecret` function. This PR closes #158.

### DIFF
--- a/pkg/transform/secrets/secrets.go
+++ b/pkg/transform/secrets/secrets.go
@@ -42,11 +42,32 @@ type MetaData struct {
 	Namespace string `yaml:"namespace"`
 }
 
+// SecretType is an enumerator for secret types
+type SecretType int
+
+const (
+	// KeystoneSecretType - keystone type for Secret
+	KeystoneSecretType = iota
+	// HtpasswdSecretType - htpasswd type for Secret
+	HtpasswdSecretType
+	// LiteralSecretType - literal type for Secret
+	LiteralSecretType
+	// BasicAuthSecretType - basicauth type for Secret
+	BasicAuthSecretType
+)
+
+var typeArray = []string{
+	"KeystoneSecretType",
+	"HtpasswdSecretType",
+	"LiteralSecretType",
+	"BasicAuthSecretType",
+}
+
 // APIVersion is the apiVersion string
 var APIVersion = "v1"
 
 // GenSecret generates a secret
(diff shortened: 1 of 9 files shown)


pkg/transform/secrets/secrets.go
```diff
@@ -42,11 +42,32 @@ type MetaData struct {
 	Namespace string `yaml:"namespace"`
 }
 
+// SecretType is an enumerator for secret types
+type SecretType int
+
+const (
+	// Keystone - keystone type for Secret
+	Keystone = iota
```
eriknelson: Usually the enum's typename is postfixed to these to prevent package level name collisions, since they're a little generic without it. So `KeystoneSecretType`, `HtpasswordSecretType` etc.

Danil-Grigorev: Ok, thank you for advise, changed.

pkg/transform/secrets/secrets.go
```diff
@@ -42,11 +42,32 @@ type MetaData struct {
...
+
+const (
+	// KeystoneSecretType - keystone type for Secret
+	KeystoneSecretType = iota
+	// HtpasswdSecretType - htpasswd type for Secret
+	HtpasswdSecretType
+	// LiteralSecretType - literal type for Secret
+	LiteralSecretType
+	// BasicauthSecretType - basicauth type for Secret
+	BasicauthSecretType
```
eriknelson: Sorry I didn't catch this before you pushed that initial change, but `Basicauth` here should be `BasicAuth` with a capital A to maintain consistency with the rest of the project. Otherwise, looks good! :+1:

Danil-Grigorev: Sorry, my bad :+1:

PR #178, opened 2019-05-27
Handle env, file and inline values of a secret

alexander-demichev:
This PR should fix issue decribed here: 

> BindPassword attribute for LDAP can be a clear text value, env var or a file reference. You can see the definition below.
https://github.com/openshift/api/blob/a2de4d01c8717dc2c9fc6429080b26e9d8bb2373/osin/v1/types.go#L156
> https://github.com/openshift/api/blob/6820a28f1dcf8f036f94a56164dfaeb842e6b28a/config/v1/types.go#L119
> Right now we only handle BindPassword if it was defined as a clear text value and skipping cases when it's something else https://github.com/fusor/cpma/blob/master/pkg/transform/oauth/ldap.go#L56 . Same problem for GitHub, GitLab, Google, OpenID.
> https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#LDAPPasswordIdentityProvider
> https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#GitHub

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"github.com/fusor/cpma/pkg/io/sftpclient"
 	"path/filepath"
 
 	"github.com/fusor/cpma/pkg/env"
@@ -28,6 +29,17 @@ func (c *Config) Fetch(path string) ([]byte, error) {
 	logrus.Infof("File:loaded: %v", dst)
 
 	return f, nil
+}
+
+// Env Fetch env vars from the OCP3 cluster
+func (c *Config) Env(envVar string) (string, error) {
+	output, err := sftpclient.GetEnvVar(c.Hostname, envVar)
+	if err != nil {
+		return "", err
+	}
+	logrus.Infof("Env:loaded: %s", output)
+
+	return output, nil
 }
 
 // LoadConfig collects and stores configuration for CPMA
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,13 +31,13 @@ func (c *Config) Fetch(path string) ([]byte, error) {
 	return f, nil
 }
 
-// Env Fetch env vars from the OCP3 cluster
-func (c *Config) Env(envVar string) (string, error) {
+// FetchEnv Fetch env vars from the OCP3 cluster
+func (c *Config) FetchEnv(envVar string) (string, error) {
 	output, err := sftpclient.GetEnvVar(c.Hostname, envVar)
 	if err != nil {
 		return "", err
 	}
-	logrus.Infof("Env:loaded: %s", output)
+	logrus.Debugf("Env:loaded: %s", envVar)
 
 	return output, nil
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"github.com/fusor/cpma/pkg/io/sftpclient"
 	"path/filepath"
 
 	"github.com/fusor/cpma/pkg/env"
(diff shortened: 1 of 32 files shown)


alexander-demichev:
@gildub I agree on coupling sftpclient and secret, it is not the best solution, but if you have any idea on getting env variables from remote host in other way let me know.

pkg/io/sftpclient/sftpclient.go
```diff
@@ -128,3 +154,19 @@ func Fetch(hostname, src, dst string) error {
...
 }
+
+// GetEnvVar get env var from remote host
+func GetEnvVar(hostname, envVar string) (string, error) {
+	session, err := NewSSHSession(hostname)
+	if err != nil {
+		return "", err
+	}
+
+	cmd := "printf $" + envVar
```
jmrodri: This is strange. Above I saw the use of `fmt.Sprintf` but not here. I would expected this to be:
```golang
cmd := fmt.Sprintf("print $%s", envVar)
```

jmrodri:
> About the patch itself, it seems to me the coupling between the sftpclient and the secrets is not appropriate at all. I'll be interested to see other view points as well.

How is `sftpclient` being tied to `secrets`? From what I saw in order to retrieve secrets that might be in an environment variable on a remote system you need to be able to get to the remote system and inspect its environment. My understanding is that the `sftpclient` is used to contact remote machines. Seems okay to have a `GetEnvVar` method on it. Unless there is other code that can do the same. `GetEnvVar` doesn't look secret specific to me. It's just a utility method to inspect an environment on a remote host.

Did I miss something? If so, please explain. I may not be seeing what you are seeing. Thanks.

pkg/transform/oauth/oauth.go
```diff
@@ -227,3 +229,30 @@ func validateClientData(clientID string, clientSecret configv1.StringSource) err
...
+		if err != nil {
+			return "", nil
+		}
+
+		secretFileString := strings.TrimSuffix(string(secretFileContent), "\n")
+		return secretFileString, nil
+	}
+
+	if secret.Env != "" {
+		secretEnv, err := config.Env(secret.Env)
```
gildub: `config.Env` is confusing because it's not related with tool configuration itself.

pkg/config/config.go
```diff
@@ -30,6 +31,17 @@ func (c *Config) Fetch(path string) ([]byte, error) {
 	return f, nil
 }
 
+// Env Fetch env vars from the OCP3 cluster
+func (c *Config) Env(envVar string) (string, error) {
```
gildub: This is fine for now, but somehow this is bothering me.
This is not directly related to this patch but we've been talking about removing config to env.go.
Maybe we just need placing Env somewhere else.

pkg/config/config.go
```diff
@@ -30,6 +31,17 @@ func (c *Config) Fetch(path string) ([]byte, error) {
...
+	output, err := sftpclient.GetEnvVar(c.Hostname, envVar)
+	if err != nil {
+		return "", err
+	}
+	logrus.Infof("Env:loaded: %s", output)
```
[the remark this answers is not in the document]
alexander-demichev: You are right, and I think outputing content of env variable to logs is not a good practice

pkg/io/sftpclient/sftpclient.go
alexander-demichev: I still didn't get used to `Sprintf`, I used Ruby way of concatenating strings:)

pkg/io/io.go
```diff
@@ -3,7 +3,7 @@ package io
 import (
 	"io/ioutil"
 
-	"github.com/fusor/cpma/pkg/io/sftpclient"
+	"github.com/fusor/cpma/pkg/io/remotehost"
```
jmrodri: that's reasonable.

pkg/io/io.go
```diff
@@ -21,3 +22,14 @@ var GetFile = func(host, src, target string) ([]byte, error) {
 	}
 	return f, nil
 }
+
+// FetchEnv Fetch env vars from the OCP3 cluster
```
jmrodri: I think it made sense in Config but I'm sure there's a reason for this move.

alexander-demichev: `FetchEnv ` is here because it's not directly related to application config.

Issue #167, opened 2019-05-20
Make GenYaml common function


alexander-demichev:
`GenYAML()` is repeated twice for secrets and ouath - https://github.com/fusor/cpma/blob/1aa6ac9fc61860ec87f9326cd38f339dcc748b56/pkg/transform/secrets/secrets.go#L88 https://github.com/fusor/cpma/blob/55ef73f43a10785b6a76bd3a2e07a428d25603a0/pkg/transform/oauth/oauth.go#L159

Maybe we need to make it common for both of them and later we will need `GenYAML()` for more CRs.

gildub:
Secrets are used by OAuth but they are might be used by other components. 
I think we should handle them on their own.
In such case we might actually want to retrieve the secrets before handling any other components.

gildub:
To share GenYAML is likely to involve interface{} to have a generic function, which means to switch over the possible objects passed as interfaces. 

I think it's better to have a small function for each component instead of a big GenYAML treating all the case. But that's only my viewpoint.

PR #179, opened 2019-05-27
Make GenYaml common function

alexander-demichev:


alexander-demichev:
Closes https://github.com/fusor/cpma/issues/167

pkg/transform/configmaps/configmaps_test.go
```diff
@@ -1,9 +1,11 @@
-package configmaps
+package configmaps_test
```
jmrodri: Why did you change the package? I find it easier to test when the tests are in the same package.

alexander-demichev: If test belongs to `configmaps` I am getting import cycle error in this test.
```
import cycle not allowed in test
package github.com/fusor/cpma/pkg/transform/configmaps (test)
	imports github.com/fusor/cpma/pkg/transform
	imports github.com/fusor/cpma/pkg/transform/configmaps
```

alexander-demichev: I thought about it, but I don't know what would be a good fit for `GenYAML`, maybe `utils/` ?

gildub: It makes sense to test genYAML in transform since it's part of it!

Issue #172, opened 2019-05-22
Handle all types of client secrets


alexander-demichev:
BindPassword attribute for LDAP can be a clear text value,  env var or a file reference. You can see the definition below.
https://github.com/openshift/api/blob/a2de4d01c8717dc2c9fc6429080b26e9d8bb2373/osin/v1/types.go#L156
https://github.com/openshift/api/blob/6820a28f1dcf8f036f94a56164dfaeb842e6b28a/config/v1/types.go#L119

Right now we only handle BindPassword if it was defined as a clear text value and skipping cases when it's something else https://github.com/fusor/cpma/blob/master/pkg/transform/oauth/ldap.go#L56 . Same problem for GitHub, GitLab,  Google, OpenID.

https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#LDAPPasswordIdentityProvider
https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#GitHub

alexander-demichev:
Values like client secret in github, gitlab, google and bind password in ldap can be represented as and encrypted file. https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html#master-node-configuration-passwords-and-other-data

We need to think if it’s possible to support translation of such values. Maybe there is a way to decrypt it - https://github.com/openshift/origin/blob/e5c7d40ea37921b02438787c1f28c79f844d1faf/pkg/cmd/server/admin/decrypt.go

PR #181, opened 2019-05-28
Return on error in secret creation

alexander-demichev:
Closes https://github.com/fusor/cpma/issues/172


pkg/transform/oauth/github.go
```diff
@@ -59,6 +59,9 @@ func buildGitHubIP(serializer *json.Serializer, p IdentityProvider, config *conf
 	secretName := p.Name + "-secret"
 	idP.GitHub.ClientSecret.Name = secretName
 	secretContent, err := fetchStringSource(github.ClientSecret, config)
+	if err != nil {
+		return nil, nil, nil, err
```
eriknelson: If you're returning 3 nils, I have to suggest wrapping up all the values in a type that captures the return value of a function. This isn't a blocking issue for this PR, but we should consider what this fn is actually returning and maybe abstract it.

jmrodri: Feels like buildGitHubIP should return the thing it built or an error.

PR #183, opened 2019-05-29
Handle encrypted secret values

alexander-demicev:
Values like client secret in github, gitlab, google and bind password in ldap can be represented as and encrypted file. https://docs.openshift.com/container-platform/3.11/install_config/master_node_configuration.html#master-node-configuration-passwords-and-other-data

This means that CPMA should download key and encrypted file, decrypt file content and pass decrypted value to secretgen function. All decryption logic is placed inside `fetchStringSource()` in `oauth.go`


gildub:
When I ran it outside of `$GOPATH` it seemed `cli-runtime/pkg/genericclioptions/printers/` was in conflict with another imported package.

gildub:
This is obsolete as secret are handled differently. 
If needed please create a new issue.

PR #184, opened 2019-05-29
Remove Config

jmontleon:
I have some strong feelings about config.

'config' in this new structure was always meant to be CPMA's config, not some reference to OCP configuraiton files. I feel like this line is being grossly blurred.

Furthermore our Config is storing two settings pulled from viper. Let's just use viper unless we have a need to store something in a Config that isn't coming from viper.

Even if the config.yml ends up containing an array of hosts for Source(s) there's nothing preventing us from iterating at a high level and writing configs to sub-directories based on hosts. It'd probably take < 1 hour to deal with this scenario unless I'm missing something IF we ever find we need to deal with it.

I've also rearranged decode into it's own package, as config was never meant to be related to openshift configs I'm not sure why it was there.

I've moved all the stuff for fetching strings/envs/files under IO. I feel like this can be the start of our 'FileManager' if we fill it out with file writing/reading.

@eriknelson @alexander-demichev @jmrodri @eriknelson, thoughts?


gildub:
@jmontleon,

That sounds and looks good.

The decode stuff was initially stored under config because for me config was the place to handle openshift sources and not CPMA config. FileManager seems to be the right way to clearly deal with it.

PR #185, opened 2019-05-31
Prompt missing config values

alexander-demichev:
CPMA will prompt all missing config values
demo


pkg/env/env.go
```diff
@@ -58,12 +59,65 @@ func InitConfig() error {
...
-		return errors.New("Can't read config file")
+		logrus.Debug("Can't read config file, all values will be prompted, err: ", err)
 	}
 
+	promtMissingValues()
+
 	return nil
 }
 
+func promtMissingValues() {
```
jwmatthews: Typo 'promt' should be 'prompt'
so..

```golang
func promptMissingValues() {
```

jwmatthews:
@alexander-demichev very nice, love the terminal recording with the PR, awesome.

PR #187, opened 2019-06-03
cmd Transform: InitConfig must happen first

gildub:
InitConfig() block (cmd/transform.go#L3) must happen before InitLogger because the latter uses viperConfig.


gildub:
Closing as this isn't an issue as memory for viper object is allocated in init() https://github.com/fusor/cpma/blob/master/pkg/env/env.go#L29

PR #189, opened 2019-06-03
Improve SDN report message

alexander-demichev:


pkg/transform/sdn_transform.go
```diff
@@ -60,14 +60,17 @@ func (e SDNExtraction) buildReportOutput() (Output, error) {
 		Component: SDNComponentName,
 	}
 
+	clusterNetworkComment := `Networks must be configured during installation,
```
gildub: Why not use a constant?

alexander-demichev: You are right.

PR #190, opened 2019-06-05
Get config values from CLI arguments

alexander-demichev:
This PR introduces ability to get config values from CLI arguments,  as an alternative to prompting and yaml file.
`./bin/cpma --source master-0.ademicev3112.lab.rdu2.cee.redhat.com --login root --key /Users/ademicev/.ssh/quicklab.key --port 22 -o ./data`

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -90,6 +90,7 @@ func promptMissingValues() {
 		login := ""
 		prompt := &survey.Input{
 			Message: "SSH login",
+			Default: "root",
 		}
 		survey.AskOne(prompt, &login, nil)
 		sshCreds["login"] = login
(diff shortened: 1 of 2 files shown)


pkg/env/env.go
```diff
@@ -62,6 +68,8 @@ func InitConfig() error {
 		logrus.Debug("Can't read config file, all values will be prompted, err: ", err)
 	}
 
+	getNestedArgValues()
```
arapov: IIRC, this function is Ok to implement as long as you can't do "unmarshal" from config here. We did this in past, and I don't have a time to look into. I hope you evaluated this option.

Otherwise, entire PR looks good to me.

alexander-demichev: We can't get nested values in the same way as `output-dir` or `source`.

PR #192, opened 2019-06-05
Ask to create config if it wasn't found

alexander-demichev:
demo


pkg/env/env.go
```diff
@@ -64,14 +65,17 @@ func InitConfig() error {
...
-		logrus.Debug("Can't read config file, all values will be prompted, err: ", err)
-	}
+	err = viperConfig.ReadInConfig()
 
 	getNestedArgValues()
 
 	promptMissingValues()
 
+	if err != nil {
+		createConfigFile()
```
arapov: @alexander-demichev I guess we will have more places similar to this one, where we use survey for query user for an input. I would suggest to use some well defined prefix for such queries. So that we know exactly, by the name of the function, that we're asking user. E.g. use "cli prefix" ... cliCreateConfig().

PR #193, opened 2019-06-05
add etcd report

jmontleon:


jmontleon:
The Cluster ETCD Client Port and TLS Cipher Suite Configuration were two things called out as important for us to check (and ideally provide configuration for, however that is not possible at this time)

jmontleon:
@alexander-demichev thanks for the review! Can you give it one more quick once over. I had to update the PR for a go.mod conflict. Shouldn't be anything else changed.

PR #194, opened 2019-06-06
Add cluster discovery feature

alexander-demichev:
demo


jwmatthews:
Excellent, this is exactly what I envisioned from day-1 of the tool.

Issue #203, opened 2019-06-07
Allow insecure host key


alexander-demichev:
Prompt to allow insecure host key

Using insecure key could be done by using HostKeyCallback: ssh.InsecureIgnoreHostKey() here ->
https://github.com/fusor/cpma/blob/master/pkg/io/remotehost/remotehost.go#L56

PR #204, opened 2019-06-10
Allow insecure host key

alexander-demichev:
Closes https://github.com/fusor/cpma/issues/203

I didn't include insecure key, because it's mainly for our dev purpose. 
@jmontleon Can you check if it works for you?

### DIFF
--- a/examples/cpma-config.example.yaml
+++ b/examples/cpma-config.example.yaml
@@ -4,6 +4,7 @@ SSHCreds:
   Login: "root"
   PrivateKey: "/home/example/.ssh/key"
   Port: 22
+InsecureHostKey: true
 OutputDir: "./data"
 # MasterConfigFile and NodeConfigFile are optional fields
 # Use only if cluster was configured with different config locations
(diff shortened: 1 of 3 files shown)


examples/cpma-config.example.yaml
```diff
@@ -4,6 +4,7 @@ SSHCreds:
   Login: "root"
   PrivateKey: "/home/example/.ssh/key"
   Port: 22
+InsecureKey: true
```
arapov: This should be InsecureHostKey, to comment itself. Or, perhaps, even better to keep it inline with ssh go package... InsecureIgnoreHostKey.

alexander-demichev: Inlining it with ssh credentials may require some additional and unnecessary engineering. `InsecureHostKey` is a boolean and we are reading ssh config as `map[string]string`  + more changes in `root.cmd` because this flag doesn't have string value

arapov: By keep it inline I meant to keep the same name. InsecureKey doesn't tell you much, unlike InsecureHostKey.

jmontleon:
Yes, this works for me when I haven't already SSH'd into the host using the hostname specified. Thanks!

Issue #197, opened 2019-06-06
OAuth: Add missing and report unsupported parameters 


gildub:
Add the following parameters which are supported in OCP4 OAuth CRD spec:
- template -> error -> name
- template -> login -> name
- template -> providerSelection -> name
- tokenConfig -> accessTokenMaxAgeSeconds

Mark as unsupported, only if value is set:
- AlwaysShowProviderSelection
- AssetPublicURL
- MasterCA
- MasterPublicURl
- MasterURL
- SessionConfig:
  - sessionMaxAgeSeconds
  - sessionName
  - sessionSecretsFile
- tokenConfig -> authorizeTokenMaxAgeSeconds

jmontleon:
I think tokenConfig -> accessTokenMaxAgeSeconds was one of the items we were asked for. Would be great if we can do this.

alexander-demichev:
I will work on this

PR #207, opened 2019-06-10
Add translate and report for oAuth token config

alexander-demichev:
We can translate `accessTokenMaxAgeSeconds` in token config and append it to oAuthCR, 
but `authorizeTokenMaxAgeSeconds` became unavailable for user.

https://github.com/openshift/cluster-authentication-operator/blob/6ae1bd829bfeccc2b5beafa98315211c8deb2da0/pkg/operator2/oauth.go#L138
https://docs.openshift.com/container-platform/3.11/install_config/configuring_authentication.html#token-options


alexander-demichev:
https://github.com/fusor/cpma/issues/197

jmontleon:
Report output looks sane:

```

PR #208, opened 2019-06-10
add crio transform

jmontleon:
crio.conf is unlikely to exist in many clusters as docker was the default runtime, so it's very likely unless you add a crio.conf that you'll see a warning that the file cannot be downloaded.

If it does exist however normal processing should occur.

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -52,6 +52,7 @@ func InitConfig() error {
 	}
 	viperConfig.Set("home", home)
 
+	viperConfig.SetDefault("CrioConfigFile", "/etc/crio/crio.conf")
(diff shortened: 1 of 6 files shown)


jmontleon:
Example:
https://github.com/openshift/machine-config-operator/blob/d2c44d7cc447868030df413c04e6fc4b22f4eb90/examples/containerruntimeconfig.crd.yaml

jmontleon:
@arapov thank you for catching the commented stuff. I'll fix that.

I also added tests for the reports. I know we are falling behind on that. I'll work on going back and adding for etcd, api, sdn, and oauth as time permits this sprint.

alexander-demichev:
Looks good, thanks!

PR #212, opened 2019-06-11
Add token inactivity timeout to translate/report

alexander-demichev:


pkg/transform/oauth/oauth_test.go
```diff
@@ -58,7 +58,7 @@ func TestTransformMasterConfig(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			oauthResources, err := oauth.Translate(identityProviders, oauth.TokenConfig{AccessTokenMaxAgeSeconds: 42000, AuthorizeTokenMaxAgeSeconds: 42000})
+			oauthResources, err := oauth.Translate(identityProviders, oauth.TokenConfig{AccessTokenMaxAgeSeconds: 42000, AuthorizeTokenMaxAgeSeconds: 42000, AccessTokenInactivityTimeoutSeconds: 42000})
```
arapov: Didn't we have a conversation to keep lines no longer than 100 or 120? I think it is good thing to start doing.

alexander-demichev:
I just noticed that AccessTokenInactivityTimeoutSeconds got removed in OCP 4.1. No need to translate it, I will close this PR.

arapov:
> I just noticed that AccessTokenInactivityTimeoutSeconds got removed in OCP 4.1.
Woah!

pkg/transform/oauth/oauth_test.go
jmontleon: I'm not too picky on this subject. I'm also generally working with pretty wide monitors, so 120 would be fine with me. If others are working with less real estate and would prefer less that's fine too.

PR #213, opened 2019-06-11
Use constants instead of repeating strings everywhere

jmontleon:


pkg/transform/transform.go
```diff
@@ -93,7 +84,6 @@ func Start() {
 
 	runner.Transform([]Transform{
 		APITransform{},
-		CrioTransform{},
```
arapov: PR is straightforward, but this line. I hope it is intended change and the line was not removed accidentally.

jmontleon: Nope, it was removed by accident. I just noticed it myself and corrected it. Thank you for catching it!

PR #218, opened 2019-06-12
Improve report messages

alexander-demichev:
This PR improves report, in my opinion some parts of report don't properly match master config yaml.

I think if yaml looks like this:
```
tokenConfig:
  accessTokenMaxAgeSeconds: 123
  authorizeTokenMaxAgeSeconds: 123
```

`Name` field in report should match `accessTokenMaxAgeSeconds` and `Kind` should be `tokenConfig` so report json will look like this:
```
{
	"name": "AccessTokenMaxAgeSeconds",
	"kind": "TokenConfig",
	"supported": true,
	"confidence": "green",
	"comment": ""
},
{
	"name": "AuthorizeTokenMaxAgeSeconds",
	"kind": "TokenConfig",
	"supported": false,
	"confidence": "red",
	"comment": "Translation of AuthorizeTokenMaxAgeSeconds is not supported, it's value is 5 minutes in OCP4"
}
```

### DIFF
--- a/pkg/transform/oauth_transform.go
+++ b/pkg/transform/oauth_transform.go
@@ -3,6 +3,7 @@ package transform
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 
 	"github.com/fusor/cpma/pkg/decode"
 	"github.com/fusor/cpma/pkg/env"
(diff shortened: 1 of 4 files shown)


gildub:
This is good because it's more consistent across the assistant.
Now I wonder if we should report the data value when it makes sense so we don't lose that information and so the user can see the data that won't be taken in account.

gildub:
Well, the comment field can effectively be used for passing the value back to the user. And if needed we can add a dedicated field later on.

arapov:
I can't comment on the yaml validity for OCP4. Though the PR is clean and looks good.

@alexander-demichev, did you actually try to test this yaml output to configure OCP4 cluster?

PR #220, opened 2019-06-12
Rename configv1 to legacyconfigv1

gildub:
The main reason for renaming the package identifier is to avoid future confusion when importing the config/v1.


arapov:
Yikes! Finally!

PR #222, opened 2019-06-12
use ghodds yaml

jmontleon:
This is a first step that will allow us to start using the openshift v1 structs and remove the ones we've created by hand. 

Those structs use json tags so we need to use a yaml library that will respect those instead of yaml tags since we need to print yaml


jmontleon:
I need to edit one more file before all tests will pass. Should have it ready to go shortly.

jmontleon:
Tests should pass now.

I added a commit for the Registries transform demonstrating how to replace the custom structs with the openshift/api/config/v1 struct

PR #226, opened 2019-06-13
Init k8s api client in a separate package

alexander-demichev:
Initializing k8s api client in separate package should help us reusing code, later we'll have other parts of CPMA that require api usage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,9 @@ matrix:
 before_install:
   - env GO111MODULE=off go get -u golang.org/x/lint/golint
 
+before_script:
+  - mkdir $HOME/.kube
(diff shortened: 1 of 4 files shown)


arapov:
This looks like a WiP, and I like it. I'd like @jmontleon and @gildub to ACK it too before we merge it.

jmontleon:
I like the idea of it being reusable. I think we'll also need to get more information from the API eventually.

My only question is,  do you think if we're making this reusable for transforms, etc. that it still belongs under env or maybe just pkg/api/api.go?

alexander-demichev:
If `api` package should be usable in multiple packages, not only in `env`, then it makes sense to not have as a subpackage.

PR #231, opened 2019-06-14
Cluster report

alexander-demichev:
This PR is introducing initial steps on cluster report.  Cluster report goes through important k8s resources that will be migrated and collects information about them, in order to get a better picture of cluster.

Sample output:
```
{
  "namespaces": [
    {
      "name": "default",
      "pods": [
        {
          "name": "docker-registry-1-qch6d"
        },
        {
          "name": "registry-console-1-xsn2c"
        },
        {
          "name": "router-1-4nznn"
        }
      ]
    }
  ],
  "pvs": [
    {
      "name": "pv1"
    }
  ]
}
```


jmontleon:
Not sure if this is something we can fix.

First run resulted in this:

alexander-demichev:
@jmontleon Your output is empty:)

jmontleon:
Sorry, trying to get a better handle on what I'm seeing. The comment should be there now.

jmontleon:
Question about the structure. Do you feel this needs to be outside of transform? It feels to me like clusterreport could just be another transform pulling from the API and wherever else and dumping out to a report.

And do you see a reason why we wouldn't keep it in the report.json? If realize it doesn't look very much like the other component reports and we could rework the json output a bit if we need to in order to make it more sane. It just feels like we could include it all in one file. 

Maybe something like:

```

**Comment by jmontleon** [comment]
Other than those comments it looks like a solid start!

**Comment by jmontleon** [comment]
I notice in my old config settings were uppercased
```
---
Source: "192.168.122.30.nip.io"
SSHCreds:
```

Now they're all lower case. No problem if that was intentional, just pointing it out in case it has something to do with the errors. I also keep getting prompted for output dir every run now and it panics in the same way.

**Comment by alexander-demichev** [comment]
@jmontleon I will reply in one long-read comment:)

1. The reason you get SEGV is because `client-go` receives empty config(https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L335), this is weird behavior, I would expect them have some check there. We also miss adding `ClusterName`  to generated config. 

2. If you create config using `survey` it will be lowercased, because `viper` converts all keys to lowercase(https://github.com/spf13/viper/blob/ad5ed02fa46d33a367ae1e563a3579f1a42e867e/viper.go#L1221). Maybe we should use lowercase in example config.
Btw, my config looks like this:
```
---
Source: "master-0.example.com"
SSHCreds:
  Login: "root"
  PrivateKey: /Users/ademicev/.ssh/quicklab.key
  Port: 22
ClusterName: openshift-example-com:443
OutputDir: "./data"
```
3. I prefer keeping `cluster report` logic separately from transform, because transform is about fetching  and translating master config, `cluster report` package pulls data from API.  Having both reports in one json is totally fine. I wanted to start work on cluster report with this PR and keep it simple. We can collect outputs from both transform/report and put the in a single json.

File: pkg/api/initclient.go [hunk 1]
```diff
@@ -74,10 +81,21 @@ func CreateAPIClient(contextCluster string) error {
...
+	if Client, err = kubernetes.NewForConfig(config); err != nil {
 		return err
 	}
 
+	logrus.Debugf("API client initialized for %s", contextCluster)
+
 	return nil
 }
+
+func validateConfig(contextCluster string) error {
```
alexander-demichev: This validator should make api client initialization safe from SEGV

jmontleon:
I don't really agree with making exceptions to the ETL flow.

- The idea is extract information from the openshift 3 cluster (whether from configuration files or API)
- Transform that data into useful outputs. Report, Manifests, or whatever other useful outputs we need
- Write that out to file

I don't see how the cluster report fails to fit that scenario.

eriknelson:
> I prefer keeping cluster report logic separately from transform, because transform is about fetching and translating master config, cluster report package pulls data from API.

The point of the abstraction is that it doesn't matter where the data is coming from. You are generically extracting some data, transforming it, and outputting it somehow, as @jmontleon describes. I don't agree that it should be kept separate, it's just another transform, unless I'm missing a technical reason.

alexander-demichev:
I moved cluster report to transform, but still haven't figured out how to merge them into a single json. 
I will be happy see your feedback!

.gitignore
```diff
@@ -12,6 +12,7 @@
 *.swo
 *.swp
 cpma
+cpma.yaml
```
gildub: +1

pkg/transform/cluster_report_transform.go
```diff
@@ -0,0 +1,154 @@
...
+			StorageClass: pv.Spec.StorageClassName,
+		}
+
+		clusterReport.PVs = append(clusterReport.PVs, *reportedPV)
+	}
+}
+
+func (clusterReport *ClusterReport) reportStorageClasses(apiResources api.Resources) {
+	logrus.Debug("ClusterReport::ReportStorageClasses")
+	// Go through all storage classes and save required information to report
```
gildub: I think the comment is not needed as code explicit.

pkg/transform/transform.go
```diff
@@ -84,6 +84,7 @@ func Start() {
 
 	runner.Transform([]Transform{
 		APITransform{},
+		ClusterReportTransform{},
```
gildub: Not a biggie, but what about using just `ClusterTransform{}` although it does produce only report it still goes through the ETL flow ?

Issue #228, opened 2019-06-14
Tests: Load expected CRs and Reports from YAML and JSON files


gildub:
Currently the expected CRs and reports are mostly generated from struct.
By loading them from YAML and JSON files would save code and potentially make tests easier to maintain.

gildub:
Tests are loading configuration and expectations from file for most of the transform and this is the way to go forward.

PR #233, opened 2019-06-17
SDN test: loading expected CR and Report from YAML/JSON

gildub:
https://github.com/fusor/cpma/issues/228


alexander-demichev:
This looks simpler than before and does the same test, ACK. Thanks!

PR #235, opened 2019-06-17
Oauth test load

gildub:
https://github.com/fusor/cpma/issues/228


alexander-demichev:
Looks good, I like this change. Will also be nice to get feedback from other folks.

PR #244, opened 2019-06-19
Add transform for MasterConfig.ProjectConfig

gildub:


pkg/transform/project_transform.go
```diff
@@ -0,0 +1,157 @@
...
+			Comment:    "Not supported in OCP4: " + strconv.Itoa(e.ProjectConfig.SecurityAllocator.MCSLabelsPerProject),
+		})
+
+	reportOutput.Reports = append(reportOutput.Reports,
+		Report{
+			Name:       "SecurityAllocator.uidAllocatorRange",
+			Kind:       "ProjectConfig",
+			Supported:  false,
+			Confidence: NoConfidence,
+			Comment:    "Not supported in OCP4: " + string(e.ProjectConfig.SecurityAllocator.UIDAllocatorRange),
```
alexander-demichev: Golang way of concatenating strings is `sprintf`, but that also looks ok for me after working with `Ruby`

pkg/transform/project/project.go
```diff
@@ -0,0 +1,36 @@
...
+	if projectConfig.ProjectRequestTemplate != "" {
+		projectCR.Spec.ProjectRequestTemplate.Name = projectConfig.ProjectRequestTemplate
+	}
+
+	return &projectCR, nil
+}
+
+// Validate registry data collected from an OCP3 cluster
+func Validate(e legacyconfigv1.ProjectConfig) error {
+	return nil
```
alexander-demichev: Is there any reason to skip validation?

pkg/transform/project_transform.go
gildub: I remember reading about such practice, but you're right `Ruby` habit has taken over!
I've changed it. Thank you.

pkg/transform/project/project.go
gildub: All the fields are optional so there is nothing much to enforce.

PR #263, opened 2019-06-25
Transform: Adds scheduler component

gildub:
This complement the master configuration with Project Configuration DefaultNodeSelector which is part of scheduler in OCP4


arapov:
@gildub this patch looks good, though it wants to have tests fixed.

pkg/utils/test/load_test_data.go
```diff
@@ -11,6 +11,23 @@ import (
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
+// LoadMasterConfig loads MasterConfig
+func LoadMasterConfig(file string) (*legacyconfigv1.MasterConfig, error) {
```
arapov: @Danil-Grigorev has pointed to this piece. It looks OK to do, but @gildub is there other places to use LoadMasterConfig in the code?

gildub: Yes definitely, that was my plan in following patch/PR to push it up.

arapov: Awe! :)

PR #265, opened 2019-06-26
K8s API client moved into pkg/apiclient/k8s

gildub:
The initial driver is to structure so we can add OpenShift API client.


gildub:
@arapov, you're right, once starting working on the openshift client that it made more sense to have them together.
Thanks.
Closing this one.

arapov:
@gildub, I think we want to re-evaluate this code and PR. See @eriknelson comments in Slack.

PR #266, opened 2019-06-26
Extended cluster report

### LINKED ISSUE #251: Cluster Report Improvements
For Persistent Volumes:
- [x] Add Capacity if not already there
- [x] Add Phase (Bound, List, Available, etc.)
- [x] Add Driver (NFS, EBS, etc.)
 (Note from John, Assume we will need some logic to look at determine, not aware of a simple field of a 'type' we can just query on PV spec.)

- [x] Add Node List to report
- [x] CPU
- [x] Memory
- [x] Consumed Memory, if possible, did not see an API response for this
- [x] Number of Pods runing (this might be more difficult in that you may have to look at pods to determine what node they running on)

- [x] List Masters if possible. Didn't see a way to list this in the API, but we get a list somehow for the settings survey?

Danil-Grigorev:
Based on [#231](https://github.com/fusor/cpma/pull/231), closes #251.

- Added info about PV capacity, driver and phase.
- Added node list to output, which contains resource information.
- Master nodes are listed with other nodes by boolean value.

Sample output:
- Nodes: 
```
"nodes": [
   {
    "name": "ip-10-1-1-245.eu-central-1.compute.internal",
    "masterNode": true,
    "resources": {
     "cpu": "4",
     "memoryConsumed": "100Mi",
     "memoryCapacity": "16265188Ki",
     "runningPods": "30",
     "podCapacity": "250"
    }
   }
  ],
...
```
- PVs:
```
"pvs": [
   {
    "name": "pv1",
    "driver": {
     "nfs": {
      "server": "127.0.0.1",
      "path": "/var/lib/nfs/exports/pv1"
     }
    },
    "capacity": {
     "storage": "10G"
    },
    "phase": "Available"
   },
...
```

### DIFF
--- a/pkg/transform/cluster_report_transform.go
+++ b/pkg/transform/cluster_report_transform.go
@@ -1,11 +1,14 @@
 package transform
 
 import (
+	"strconv"
+
 	"github.com/fusor/cpma/pkg/api"
 	O7tapiroute "github.com/openshift/api/route/v1"
 	"github.com/sirupsen/logrus"
 
 	k8sapicore "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 // ClusterReportName is the cluster report name
@@ -28,6 +31,7 @@ func (e ClusterReportExtraction) Transform() ([]Output, error) {
 		PersistentVolumeList: e.PersistentVolumeList,
 		StorageClassList:     e.StorageClassList,
 		NamespaceMap:         e.NamespaceMap,
+		NodeList:             e.NodeList,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


pkg/transform/cluster_report_transform.go
```diff
@@ -51,6 +58,52 @@ func genClusterReport(apiResources api.Resources) (ClusterReport, error) {
...
+// reportNodes fills in information about nodes
+func (clusterReport *ClusterReport) reportNodes(apiResources api.Resources) {
+	logrus.Debug("ClusterReport::ReportNodes")
+
+	for _, node := range apiResources.NodeList.Items {
+		nodeReport := &NodeReport{
+			Name: node.ObjectMeta.Name,
+		}
+
+		isMaster, ok := node.ObjectMeta.Labels["node-role.kubernetes.io/master"]
```
arapov: this "ok" is nice, I hope it is inline with the k8s code.

pkg/transform/cluster_report_transform.go
```diff
@@ -51,6 +58,52 @@ func genClusterReport(apiResources api.Resources) (ClusterReport, error) {
...
+
+	runningPods := 0
+	for _, resources := range apiResources.NamespaceMap {
+		for _, pod := range resources.PodList.Items {
+			if pod.Spec.NodeName == repotedNode.Name {
+				runningPods++
+			}
+		}
+	}
+	repotedNode.Resources.RunningPods = strconv.Itoa(runningPods) + "/" + nodeStatus.Capacity.Pods().String()
```
arapov: @Danil-Grigorev what is the reason to have the string field in structure when you convert int to string here? Isn't it better to have RunningPod defined as int in the structure? May be I'm missing the story of your inspiration here, if so - share!

pkg/transform/report_output.go
```diff
@@ -13,6 +14,14 @@ type ReportOutput struct {
 	ComponentReports []ComponentReport `json:"components"`
 }
 
+// NodeResources represents a json report of Node resources
+type NodeResources struct {
+	CPU            string `json:"cpu"`
+	MemoryConsumed string `json:"memoryConsumed"`
+	MemoryCapacity string `json:"memoryCapacity"`
+	RunningPods    string `json:"runningPods"`
```
arapov: This place is mentioned above. I didn't check the rest fields, those may have the similar issue.

pkg/transform/cluster_report_transform.go
Danil-Grigorev: I'll just quote Jason on this topic:

'There was a statement made that in some places IT sets up a cluster and lets the devs at it. They start spinning things up and IT really has no idea what's where.'

'so the capacity stuff, routes exposed, is to try and give them a fuller view of what's happened since they set it up'

'how many pv's totalling how much (at least potential storage), how many nodes, how much of those nodes are being utilized, etc.'

Certainly, from the code perspective having explicit conversion from int to string is stupid.

But having a structure like `pod_count/max_pod_count` implies having some conversions in place.

In terms of the idea and a customer perspective this output looks clear. Report should contain human-readable values, that could be captured from the first glance.

arapov: I would agree in case the json is meant to be edited by humans. in our case it is just a mean to deliver the information to another application. I would insist on keeping this typed.

However if Jason have a reason, besides IT/humans doesn't know what is int/string in json, I am all for the support.

IOW, this reporting json output aren't meant for direct user consumption, even if - it is read only.

@jmontleon wdyt?

pkg/transform/report_output.go
Danil-Grigorev: In addition to the previous explanation: The structure does not have a straight analogy in k8s api, so we are free to fantasize here.

But I agree that there is a room for improvement here. Going to look what to change.

pkg/transform/cluster_report_transform.go
Danil-Grigorev: https://github.com/kubernetes/api/blob/9e5ffd1f1320950b238cfce291b926411f0af722/core/v1/resource.go#L38 this is k8s.

Here is a whole discussion on this topic: https://gist.github.com/posener/a303becac35835ad7bf5e15fe061893e TL;DR ok is better (ba dum tss).

Our mig-controller tool is based on this (1087 occurrences). If the k8s api don't use it, I suggest to start also. Let's open an issue on this, CPMA and mig-controller is from the same division, let's follow our big brother.

Issue #255, opened 2019-06-20
Route List Report


jmontleon:
Provide a list of routes. Each entry should  include at a minimum some basic information
- Name
- Namespaces
- URL

TLS Termination type in the case of SSL routes and the to field (for example the service the route is pointing at) maye also be useful fields.

The idea is to give users a view of what services are exposed from the cluster.

PR #267, opened 2019-06-26
Openshift API client and Routes List

gildub:
https://github.com/fusor/cpma/issues/255

### DIFF
--- a/pkg/api/openshift.go
+++ b/pkg/api/openshift.go
@@ -6,7 +6,6 @@ import (
 	authv1 "github.com/openshift/client-go/authorization/clientset/versioned/typed/authorization/v1"
 	routev1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/rest"
 )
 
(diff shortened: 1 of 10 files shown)


arapov:
this WiP utterly needs an examples of use. :)

alexander-demichev:
One question here, are you planning to replace k8s api with openshift? I suppose openshift client-go should contain all k8s resources + ocp specific.

gildub:
@alexander-demichev, no k8s and openshift are complementary.

pkg/api/openshift.go
```diff
@@ -0,0 +1,55 @@
...
+	Openshift sync.Once
+}
+
+// Openshift - Create a new openshift client if needed, returns reference
+func Openshift(config *rest.Config) (*OpenshiftClient, error) {
+	errMsg := "Something went wrong while initializing openshift client!\n"
+	once.Openshift.Do(func() {
+		client, err := newOpenshift(config)
+		if err != nil {
+			log.Fatal(errMsg)
```
alexander-demichev: Shouldn't it return on error?

gildub: Yes, you're right that's residual code from previous work.

gildub: Yes, it will. Thanks

gildub: @alexander-demichev, actually if Openshift creation fails it's fatal.

alexander-demichev: Hmm, shouldn't we skip report of API resources in this case? If not, then we also need to panic in k8s client initialization.

gildub: Yeah I suppose there are 2 different approaches, whether we trigger panic or escalate.
I initially thought that not being able to create an o7t client was a blocker and so is k8s because the report does rely on the APIs. 
Let's just escalate for now, I created issue https://github.com/fusor/cpma/issues/272 
to allow us to move on as we need o7t.

PR #268, opened 2019-06-27
Added report for pod resources, related to namespaces.

### LINKED ISSUE #253: Namespace Rank
Add a field to the Namespace report with quantity 

This will involve some research to determine to what extent this is possible. 

It may be that the best we can do is see what's requested by pods in the namespace. It looks like pods can be set up without resource requests in which case it just shows `resources: {}`

```
resources:
  requests: 
    cpu: "100m"
    memory: "256Mi"
```

This may involve looking at the pods to determine 
- [ ] CPU
- [ ] Memory
- [ ] Pods (Quantity)

Danil-Grigorev:
Closes #253. Every namespace has a report of used resources.

```
"namespaces": [
   {
    "name": "default",
    "latestChange": "2019-06-28T23:47:07Z",
    "resources": {
     "containerCount": 4,
     "cpuSummary": "3",
     "memSummary": "640Mi"
    },
...
```

### DIFF
--- a/pkg/transform/cluster_report_transform.go
+++ b/pkg/transform/cluster_report_transform.go
@@ -25,15 +25,12 @@ type ClusterTransform struct {
 func (e ClusterReportExtraction) Transform() ([]Output, error) {
 	logrus.Info("ClusterTransform::Transform")
 
-	clusterReport, err := genClusterReport(api.Resources{
+	clusterReport := genClusterReport(api.Resources{
 		PersistentVolumeList: e.PersistentVolumeList,
 		StorageClassList:     e.StorageClassList,
 		NamespaceMap:         e.NamespaceMap,
 		NodeList:             e.NodeList,
 	})
-	if err != nil {
-		return nil, err
-	}
 
 	output := ReportOutput{
 		ClusterReport: clusterReport,
@@ -44,7 +41,7 @@ func (e ClusterReportExtraction) Transform() ([]Output, error) {
 }
 
 // genClusterReport inserts report values into structures for json output
-func genClusterReport(apiResources api.Resources) (clusterReport ClusterReport, err error) {
+func genClusterReport(apiResources api.Resources) (clusterReport ClusterReport) {
 	clusterReport.reportNodes(apiResources)
 	clusterReport.reportNamespaces(apiResources)
 	clusterReport.reportPVs(apiResources)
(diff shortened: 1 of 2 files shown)


pkg/transform/cluster_report_transform.go
```diff
@@ -35,20 +35,19 @@ func (e ClusterReportExtraction) Transform() ([]Output, error) {
...
 
-	outputs := []Output{output}
-	return outputs, nil
+	outputs = []Output{output}
+	return
 }
 
-func genClusterReport(apiResources api.Resources) (ClusterReport, error) {
-	clusterReport := ClusterReport{}
+func genClusterReport(apiResources api.Resources) (clusterReport ClusterReport, err error) {
```
gildub: Maybe more empty lines in this block?

pkg/transform/cluster_report_transform.go
```diff
@@ -148,6 +155,4 @@ func (e ClusterTransform) Extract() (Extraction, error) {
 }
 
 // Name returns a human readable name for the transform
-func (e ClusterTransform) Name() string {
-	return ClusterReportName
-}
+func (e ClusterTransform) Name() (ClusterReportName string) { return }
```
gildub: This is not the equivalent.
The new Name() returns an empty value no matter what

Danil-Grigorev: I'm going to add the tests, you are correct, sorry for the mess.

pkg/transform/cluster_report_transform.go
gildub: Actually never mind my comment, the code I'm talking about is legacy (existed before this patch).

pkg/transform/report_output.go
```diff
@@ -69,6 +72,13 @@ type RouteReport struct {
 	WildcardPolicy    O7tapiroute.WildcardPolicyType     `json:"wildcardPolicy"`
 }
 
+// ContainerResourcesReport represents json report for aggregated container resources
+type ContainerResourcesReport struct {
+	ContainerCount int64              `json:"containerCount"`
```
alexander-demichev: Is there any specific reason for int64?

pkg/transform/report_output.go
```diff
@@ -69,6 +72,13 @@ type RouteReport struct {
...
+	CPUsummary     *resource.Quantity `json:"cpuSummary"`
+	MemSummary     *resource.Quantity `json:"memSummary"`
```
alexander-demichev: I think MemTotal and CPUTotal are appropriate names for these cases.

alexander-demichev:
@gildub Named return values look strange, but I think it is ok since golang doesn't have strong convention about things like `new(Struct)` vs `&Struct{}`. Before golang get customizable configuration for linter, like Ruby's rubocop, we can't forbid usage of named returns.

pkg/transform/report_output.go
```diff
@@ -69,6 +72,13 @@ type RouteReport struct {
 	WildcardPolicy    O7tapiroute.WildcardPolicyType     `json:"wildcardPolicy"`
 }
 
+// ContainerResourcesReport represents json report for aggregated container resources
+type ContainerResourcesReport struct {
+	ContainerCount int64              `json:"containerCount"`
+	CPUsummary     *resource.Quantity `json:"cpuSummary"`
```
alexander-demichev: What is the need of having `CPUsummary ` as resource.Quantity in final report? Why not store it string?

Issue #262, opened 2019-06-24
RFC: clean-up cpma parameters


arapov:
1. do we actually need *both* --cluster-name and --source? 
1.a. if yes - parameters are crying for better names
1.b. even code suggests they are the same
```
env.Config().BindPFlag("Source", rootCmd.PersistentFlags().Lookup("source"))
env.Config().BindPFlag("ClusterName", rootCmd.PersistentFlags().Lookup("source"))
```
2. I don't think we should keep "OCP3" in help, we aren't going to have OCP4 in future, iiuc.
2.a. instead we OCP3 message must be stressed right before the list of parameters.
3. suggestion is to rearrange and change parameters:

```
OCP3 migration bla bla bla....
    --config string        config file (default searches: ./cpma.yaml, $HOME/cpma.yml)
-h, --hostname string      cluster hostname
-p, --ssh-port string      ssh port cluster ssh port   (default: 22)
-k, --ssh-keyfile string   ssh keyfile path            (default: $HOME/.ssh/id_rsa)
-l, --ssh-login string     ssh login                   (default: root)
-i, --allow-insecure-host  allow insecure ssh host key << give better describtion in help >>
-o, --output-dir string    set the directory to store extracted configuration. (default: current directory)
-v, --verbose              verbose output
-d, --debug                debug output
-h, --help                 help for cpma
```

Please, +1 or comment!

PR #273, opened 2019-07-01
Clean up parameters

alexander-demichev:
Closes https://github.com/fusor/cpma/issues/262


alexander-demichev:
I updated readme with last force push

Issue #278, opened 2019-07-02
Decouple ETL from configuration file fetching


gildub:
Because of possible use cases where SSH restrictions won't allow fetching the configuration files needed by the ETL we need have a simpler/clearer workflow for the user to launch CPMA.

We already locally cache files fetched remotely and the user can retrieve (by script or not) the required files prior to launching the tool.

A better workflow would be:
CPMA starts:
- If a default config isn't avail, ask user: remote or local?, output dir?, configuration files path/name if different than default. and hostname (because that's where files are stored locally even when fetched ), store config?
- If remote, ask user, for remote user, port, ssh key, store in the config if avail.
- Starts ETL

PR #279, opened 2019-07-03
Use local config files

alexander-demichev:
Closes https://github.com/fusor/cpma/issues/278

If ssh is not available, we ask user for source of config files, local or remote. If remote then we keep them same workflow we have right now, if local then just use survey to get paths listed here.
This PR also adds ability for user to choose how API client should be created, using cluster name from current context or prompt.
 
I will try to add tests for `survey` inputs later and update README after review.

Local files as source:
local
Remote files:
remote

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -15,8 +15,6 @@
 package cmd
 
 import (
-	"path"
-
 	"github.com/fusor/cpma/pkg/env"
 	"github.com/fusor/cpma/pkg/transform"
 	"github.com/sirupsen/logrus"
@@ -55,7 +53,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&env.Port, "ssh-port", "p", "", "OCP3 ssh port")
 
 	// Get config file from CLI argument an save to viper config
-	rootCmd.PersistentFlags().StringP("output-dir", "o", path.Dir("."), "set the directory to store extracted configuration.")
+	rootCmd.PersistentFlags().StringP("output-dir", "o", "", "set the directory to store extracted configuration.")
 	env.Config().BindPFlag("OutputDir", rootCmd.PersistentFlags().Lookup("output-dir"))
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -55,6 +55,30 @@ func init() {
 	// Get config file from CLI argument an save to viper config
 	rootCmd.PersistentFlags().StringP("output-dir", "o", "", "set the directory to store extracted configuration.")
 	env.Config().BindPFlag("OutputDir", rootCmd.PersistentFlags().Lookup("output-dir"))
+
+	// Set log level from CLI argument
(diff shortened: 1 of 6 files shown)


pkg/env/env.go
```diff
@@ -71,55 +65,117 @@ func InitConfig() error {
...
-			case err.Error() == "interrupt":
-				return errors.Wrap(err, "ctrl-C pressed. Exiting.")
-			default:
-				return errors.Wrap(err, "Error in creating config file")
-			}
+			return err
 		}
 
-		logrus.Debug("Can't read config file, all values were prompted and new config was asked to be created, err: ", readConfigErr)
+		viperConfig.Set("OutputDir", outPutDir)
```
arapov: a minor thing, worth to fix later: capital P in output looks really weird! :)

pkg/env/env.go
```diff
@@ -264,6 +396,15 @@ func surveyCreateConfigFile() error {
 	return nil
 }
 
+func handleInterrupt(err error) error {
+	switch {
+	case err.Error() == "interrupt":
+		return errors.Wrap(err, "ctrl-C pressed. Exiting.")
```
arapov: All you want to write here is "Exiting", terminal will show ^C itself.

cmd/root.go
```diff
@@ -55,8 +53,32 @@ func init() {
 	rootCmd.PersistentFlags().StringVarP(&env.Port, "ssh-port", "p", "", "OCP3 ssh port")
 
 	// Get config file from CLI argument an save to viper config
-	rootCmd.PersistentFlags().StringP("output-dir", "o", path.Dir("."), "set the directory to store extracted configuration.")
+	rootCmd.PersistentFlags().StringP("output-dir", "o", "", "set the directory to store extracted configuration.")
```
gildub: Why changing back to ""? Because "." fixes the issue to save current dir as default `OutputDir`

cmd/root.go
```diff
@@ -55,8 +53,32 @@ func init() {
...
 	env.Config().BindPFlag("OutputDir", rootCmd.PersistentFlags().Lookup("output-dir"))
+
+	// Set log level from CLI argument
+	rootCmd.PersistentFlags().String("config-source", "", "source for OCP3 config files, accepted values: remote or local")
+	env.Config().BindPFlag("ConfigSource", rootCmd.PersistentFlags().Lookup("config-source"))
+
+	// Get crio config file location
+	rootCmd.PersistentFlags().String("crio-config", "", "path to crio config file")
```
gildub: Can we make the default OCP 3.x values the default value for CPMA?

Same for the other ones: master-config, node-config, etc.

alexander-demichev: Yes, default values are set only in case of usage ssh as source for config

alexander-demichev: https://github.com/fusor/cpma/pull/279/files#diff-1e94282afa1637126c37c2d26509a2d4R259

cmd/root.go
alexander-demichev: +1

PR #283, opened 2019-07-04
Config arguments rearrangement

Danil-Grigorev:
- Fixed logic in retrival of config variables.
- Removed nested ssh values.
- Now every variable from config could be defined as following.

| ENV variable               | Config equivalent    |
|----------------------------|----------------------|
| CPMA_CLUSTERNAME           | clustername          |
| CPMA_SSHPORT               | sshport              |
| CPMA_CREATECONFIG          | createconfig         |
| CPMA_OUTPUTDIR             | outputdir            |
| CPMA_SSHPRIVATEKEY         | sshprivatekey        |
| CPMA_LOGIN                 | login                |
| CPMA_CONFIGSOURCE          | configsource         |
| CPMA_PASSWD                | passwd               |
| CPMA_HOSTNAME              | hostname             |
| CPMA_SSHLOGIN              | sshlogin             |
| CPMA_CRIOCONFIGFILE        | crioconfigfile       |
| CPMA_ETCDCONFIGFILE        | etcdconfigfile       |
| CPMA_REGISTRIESCONFIGFILE  | registriesconfigfile |
| CPMA_HOME                  | home                 |
| CPMA_MASTERCONFIGFILE      | masterconfigfile     |
| CPMA_NODECONFIGFILE        | nodeconfigfile       |

Other environment variables:

| ENV variable               |  Description      				 |
|----------------------------|---------------------------------------------------|
| CPMA_CREATECONFIG          |  Enables or disables authomatic config creation   |

Other changes:

Removed nested ssh variables
```
SSHCreds:
  Login: "root"
  PrivateKey: "/home/example/.ssh/key"
  Port: 22
```
substituted with
```
sshlogin: root
sshport: "22"
sshprivatekey: "/home/example/.ssh/key"
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -48,9 +48,14 @@ func init() {
 	env.Config().BindPFlag("ClusterName", rootCmd.PersistentFlags().Lookup("cluster-name"))
 
 	// Get ssh config values from CLI argument
-	rootCmd.PersistentFlags().StringVarP(&env.Login, "ssh-login", "l", "", "OCP3 ssh login")
-	rootCmd.PersistentFlags().StringVarP(&env.PrivateKey, "ssh-keyfile", "k", "", "OCP3 ssh keyfile path")
-	rootCmd.PersistentFlags().StringVarP(&env.Port, "ssh-port", "p", "", "OCP3 ssh port")
+	rootCmd.PersistentFlags().StringP("ssh-login", "l", "", "OCP3 ssh login")
+	env.Config().BindPFlag("SSHLogin", rootCmd.PersistentFlags().Lookup("ssh-login"))
+
+	rootCmd.PersistentFlags().StringP("ssh-keyfile", "k", "", "OCP3 ssh keyfile path")
+	env.Config().BindPFlag("SSHPrivateKey", rootCmd.PersistentFlags().Lookup("ssh-keyfile"))
+
+	rootCmd.PersistentFlags().StringP("ssh-port", "p", "", "OCP3 ssh port")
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -48,9 +48,14 @@ func init() {
 	env.Config().BindPFlag("ClusterName", rootCmd.PersistentFlags().Lookup("cluster-name"))
 
 	// Get ssh config values from CLI argument
-	rootCmd.PersistentFlags().StringVarP(&env.Login, "ssh-login", "l", "", "OCP3 ssh login")
-	rootCmd.PersistentFlags().StringVarP(&env.PrivateKey, "ssh-keyfile", "k", "", "OCP3 ssh keyfile path")
-	rootCmd.PersistentFlags().StringVarP(&env.Port, "ssh-port", "p", "", "OCP3 ssh port")
+	rootCmd.PersistentFlags().StringP("ssh-login", "l", "", "OCP3 ssh login")
```
gildub: Why removing the explicit options for 'ssh-login' and others?
We're going back and forth with that, let's keep both short and long one, unless there is good reason not to.

pkg/env/env_test.go
```diff
@@ -14,7 +14,7 @@ import (
 
 func TestInitConfig(t *testing.T) {
 	t.Parallel()
-	ConfigFile = "testdata/cpma-config.yml"
+	ConfigFile = "testdata/src/cpma-remote-config.yml"
```
gildub: Why is `src` sub folder needed?

Danil-Grigorev: Good note, it is not needed any more. I just had it, when was trying to generate a `config` file, and the `out` folder will be the location for the second one for comparison. Going to fix it now.

PR #286, opened 2019-07-05
Local feature use OutputDir for application working dir

gildub:
This also renames "OutputDir" to "WorkDir".

`WorDir` is the base when working in `Local` mode in which case files will be searched under the `hostname`.

### DIFF
--- a/pkg/io/io.go
+++ b/pkg/io/io.go
@@ -60,7 +60,7 @@ func fetchFromRemote(src string) ([]byte, error) {
 }
 
 func fetchFromLocal(src string) ([]byte, error) {
-	localSrc := filepath.Join(env.Config().GetString("WorkDir"), env.Config().GetString("Source"), src)
+	localSrc := filepath.Join(env.Config().GetString("WorkDir"), env.Config().GetString("Hostname"), src)
 	f, err := ioutil.ReadFile(localSrc)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


pkg/io/io.go
```diff
@@ -65,7 +65,8 @@ func fetchFromRemote(src string) ([]byte, error) {
 }
 
 func fetchFromLocal(src string) ([]byte, error) {
-	f, err := ioutil.ReadFile(src)
+	localSrc := filepath.Join(env.Config().GetString("WorkDir"), env.Config().GetString("Source"), src)
```
alexander-demichev: "Source" is now "Hostname":) 
Does this mean that we always ask user for hostname? If path to config files is relative to `workDir`/`Hostname` then we should add default values to survey when asking for crio, etcd, etc. configs.

gildub: Oh yeah, not source anymore, ;) . Thanks!

Yes we always need to ask for Kubeconfig first and from there the cluster. If the user selects prompt then we need to obtain the equivalent.

Default values to survey are absolute path already, but relative to the hostname base, I don't think we need to do anything about that besides as long as it's clear enough in the Readme file.

alexander-demichev: Does it make sense to ask for master hostname if no ssh used later?

gildub: Well it does makes sense to ask for the name of the host the files belongs to

alexander-demichev: I get it now, in some cases we may want to store configs for multiple clusters.

Danil-Grigorev: I agree that there should be some kind of predefined location for cluster files, so the user could run the tool for multiple clusters. `Source` looks good to me.

PR #288, opened 2019-07-08
Test for io pkg

gildub:
Adds tests to cover FetchFile(fetchFromRemote and FetchFromLocal), FetchEnv and FetchStringSource

Also refactored fetchFromRemote which was (historically) fetching file from local first. This is not needed anymore since fetchFromLocal is used by FetchFile.

### DIFF
--- a/pkg/io/io.go
+++ b/pkg/io/io.go
@@ -33,35 +33,30 @@ var FetchFile = func(src string) ([]byte, error) {
 
 func fetchFromRemote(src string) ([]byte, error) {
 	dst := filepath.Join(env.Config().GetString("Hostname"), src)
-	f, err := ReadFile(dst)
+	host := env.Config().GetString("Hostname")
+
+	cmd := fmt.Sprintf("sudo cat %s", src)
+	output, err := remotehost.RunCMD(host, cmd)
 	if err != nil {
-		host := env.Config().GetString("Hostname")
-
-		cmd := fmt.Sprintf("sudo cat %s", src)
-		output, err := remotehost.RunCMD(host, cmd)
-		if err != nil {
-			return nil, err
-		}
-
-		if output == "" {
-			msg := fmt.Sprintf("Empty or missing file: %s", dst)
-			return nil, errors.New(msg)
-		}
-
-		err = WriteFile([]byte(output), dst)
-		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/io/io.go
```diff
@@ -33,35 +33,30 @@ var FetchFile = func(src string) ([]byte, error) {
 
 func fetchFromRemote(src string) ([]byte, error) {
 	dst := filepath.Join(env.Config().GetString("Hostname"), src)
-	f, err := ReadFile(dst)
-	if err != nil {
-		host := env.Config().GetString("Hostname")
+	host := env.Config().GetString("Hostname")
```
Danil-Grigorev: Just a note, may be extract `host` before creating `dst`, and then set the `dst` as `dst := filepath.Join(host, src)`? It will be more readable IMHO.

pkg/io/io_test.go
```diff
@@ -0,0 +1,111 @@
...
+		name     string
+		remote   bool
+		expected []byte
+		filename string
+	}{
+		{
+			name:     "Fetch from remote",
+			remote:   true,
+			filename: "remote-file",
+			expected: []byte{0x72, 0x65, 0x6d, 0x6f, 0x74, 0x65, 0x20, 0x76, 0x61, 0x6c, 0x75, 0x65},
```
Danil-Grigorev: Not sure that everyone will be able to comprehend the HEX codes there. Wouldn't `string` value be better?

pkg/io/io_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func TestStringSource(t *testing.T) {
+	testCases := []struct {
+		name     string
+		filename string
+		expected string
+	}{
+		{
+			name:     "Fetch String Source ENV variable",
```
Danil-Grigorev: Looks like this test is fetching source from `Value`, not from `env`. Not quite sure about that though.

pkg/io/io_test.go
gildub: Fair enough!

pkg/io/io_test.go
gildub: +1, let me clarify that!

pkg/io/io.go
gildub: Of course! Thanks

PR #292, opened 2019-07-09
Cluster report: Refactor and adding tests

alexander-demichev:
I refactored cluster report to make it more testable and added tests

### DIFF
--- a/pkg/transform/cluster/cluster.go
+++ b/pkg/transform/cluster/cluster.go
@@ -1,0 +1,241 @@
+package cluster
+
+import (
+	"github.com/fusor/cpma/pkg/api"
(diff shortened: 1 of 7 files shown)


pkg/transform/cluster/cluster.go
```diff
@@ -0,0 +1,241 @@
+package cluster
+
+import (
+	"github.com/fusor/cpma/pkg/api"
+	O7tapiroute "github.com/openshift/api/route/v1"
+	"github.com/sirupsen/logrus"
+	k8sapicore "k8s.io/api/core/v1"
```
gildub: An empty line is needed here to separate different source repos

PR #299, opened 2019-07-10
Adds image tests and adjusts decode test

gildub:


pkg/decode/decode_test.go
```diff
@@ -1,6 +1,7 @@
 package decode
 
 import (
+	"fmt"
```
Danil-Grigorev: I feel like it should not be there.

gildub: Done.

Issue #254, opened 2019-06-20
RBAC Report


jmontleon:
This is a stretch goal for the Sprint 8

Useful things to report in this category may include:

- [ ] Groups
- [ ] Users
- [ ] Roles
- [ ] Custom Clusterroles
- [ ] Custom Rolebindings
- [ ] Custom SCC’s

PR #305, opened 2019-07-11
Add RBAC report with tests

alexander-demichev:


alexander-demichev:
Closes https://github.com/fusor/cpma/issues/254

pkg/api/requests.go
```diff
@@ -1,8 +1,10 @@
 package api
 
 import (
+	o7tapiauth "github.com/openshift/api/authorization/v1"
 	o7tapiroute "github.com/openshift/api/route/v1"
-
+	o7tapisecurity "github.com/openshift/api/security/v1"
+	o7tapiuser "github.com/openshift/api/user/v1"
```
gildub: An empty line is needed to separate repos famillies

pkg/api/requests.go
```diff
@@ -15,6 +18,16 @@ type Resources struct {
 	NodeList             *k8sapicore.NodeList
 	StorageClassList     *k8sapistorage.StorageClassList
 	NamespaceList        []NamespaceResources
+	RBACResources        RBACResources
```
gildub: Just wondering why not listing Users, Groups, Roles, and SecurityContextConstraint as such instead of wrapping them in RBAC?

pkg/transform/cluster/cluster.go
```diff
@@ -2,12 +2,13 @@ package cluster
 
 import (
 	"github.com/fusor/cpma/pkg/api"
-	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/api/resource"
-
+	o7tapiauth "github.com/openshift/api/authorization/v1"
 	o7tapiroute "github.com/openshift/api/route/v1"
+	"github.com/sirupsen/logrus"
```
gildub: github.com block of imports should be separated from k8s.io and so forth

pkg/api/requests.go
alexander-demichev: I my opinion things related to RBAC should go under separate attribute for keeping obvious to which part of report user, group, etc belong. It will be easier later to parse json in UI.

gildub: I understand the choice just that I'm not sure about it.

Issue #310, opened 2019-07-16
Add ability to generate either report or CR


alexander-demichev:
One of use cases for CPMA can be using it as analytics tool only, so we want to disable CR generation sometimes. This should be done by adding new config values like `ManifestOnly` and `ReportOnly` and the default is generating both of them.

arapov:
If the case is confirmed it worth to do indeed.

PR #312, opened 2019-07-16
Make generation of manifests and report optional

alexander-demichev:
Add ability to generate either report or CR. I know this is not the best solution, but it is tested in e2e tests. We can rethink the approach after feature freeze.

Closes https://github.com/fusor/cpma/issues/310


gildub:
Per our discussion let's close this one and create a new PR.

Issue #256, opened 2019-06-20
Resource Quota Report/Transform


jmontleon:
Determine if Resource Quotas are supported on Openshift 4. Transform if it's possible and makes sense, but at least report.

gildub:
Cluster and Resource Quotas as been added to reporting by PR#308

PR #314, opened 2019-07-18
Cluster Resource Quotas and Resource Quotas manifests

gildub:
This extends cluster_transform  (thus the file rename) to generate CRs as well.

Close https://github.com/fusor/cpma/issues/256

### DIFF
--- a/pkg/transform/clusterquota/clusterquota.go
+++ b/pkg/transform/clusterquota/clusterquota.go
@@ -6,8 +6,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// BuildManifest definitions
-func BuildManifest(quota o7tapiquota.ClusterResourceQuota) (*o7tapiquota.ClusterResourceQuota, error) {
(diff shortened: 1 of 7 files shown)


pkg/transform/clusterquota/clusterquota.go
```diff
@@ -0,0 +1,25 @@
+package clusterquota
+
+import (
+	o7tapiquota "github.com/openshift/api/quota/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// BuildManifest definitions
+func BuildManifest(quota o7tapiquota.ClusterResourceQuota) (*o7tapiquota.ClusterResourceQuota, error) {
```
alexander-demichev: Similar functions for oauth and project are called `Translate`, I prefer us to keep similar names in such cases. You can rename this to `Translate` or rename other ones to` BuildManifest`

gildub: Consider it done!

PR #315, opened 2019-07-22
Options to filter manifests or reports modes

gildub:
Closes https://github.com/fusor/cpma/issues/310

### DIFF
--- a/pkg/env/env.go
+++ b/pkg/env/env.go
@@ -93,8 +93,15 @@ func setConfigLocation() (err error) {
 }
 
 func surveyMissingValues() error {
-	err := surveyConfigSource()
-	if err != nil {
+	if err := surveyManifests(); err != nil {
+		return err
+	}
+
+	if err := surveyReports(); err != nil {
+		return err
+	}
+
+	if err := surveyConfigSource(); err != nil {
 		return err
 	}
 
@@ -114,8 +121,7 @@ func surveyMissingValues() error {
 		return errors.New("Accepted values for config-source are: remote or local")
 	}
 
-	err = createAPIClients()
-	if err != nil {
+	if err := createAPIClients(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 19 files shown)


pkg/env/env.go
```diff
@@ -93,7 +93,17 @@ func setConfigLocation() (err error) {
 }
 
 func surveyMissingValues() error {
-	err := surveyConfigSource()
+	err := surveyManifests()
+	if err != nil {
+		return err
+	}
+
```
arapov: Hey, we agreed to write such statement as:
``
if err := surveyManifests(); err != nil {
    return err
}
``
a while back. Could you change these before merge please.
See https://github.com/fusor/cpma/blob/d28c4c1fd45f70af783a20488ed8865c1169b27a/pkg/env/env.go#L46 for example.

arapov:
Also, we want Travis to succeed :)

pkg/transform/api_transform_test.go
```diff
@@ -24,6 +24,7 @@ var loadAPIExtraction = func() transform.APIExtraction {
 }()
 
 func TestAPIExtractionTransform(t *testing.T) {
+	t.Skip()
```
alexander-demichev: Is this needed?

gildub: Nope that's a leftover. Thanks!

pkg/env/env.go
gildub: Yes you're right, this is actually legacy code we need to change. Thanks

pkg/env/env.go
```diff
@@ -159,6 +165,44 @@ func surveyConfigSource() error {
 	return nil
 }
 
+func surveyManifests() error {
+	manifests := viperConfig.GetString("Manifests")
+	if !viperConfig.InConfig("manifests") {
```
alexander-demichev:
```suggestion
	if !viperConfig.InConfig("manifests") && manifests == "" {
```

pkg/env/env.go
```diff
@@ -159,6 +165,44 @@ func surveyConfigSource() error {
...
+		} else {
+			viperConfig.Set("Manifests", false)
+		}
+	}
+	return nil
+}
+
+func surveyReports() error {
+	reports := viperConfig.GetString("Reports")
+	if !viperConfig.InConfig("reports") {
```
alexander-demichev:
```suggestion
	if !viperConfig.InConfig("reports") && reports == "" {
```

PR #317, opened 2019-07-23
Document manifest creation workflow

alexander-demichev:


docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
+## Generating manifests using CPMA and applying them
+
+CPMA analyzes OCP3 configuration files and when possible transforms them to CRs, generated CRs are used to configure OCP4 cluster as `day two` operations.
+
+---
+
+
+### 1. Prerequisites
+
+Prior to working with CPMA you need to deploy an OCP 3.7+ cluster(OCP 3.7, 3.9, 3.10, 3.11 will be supported).
```
gildub: I suggest s/will be supported/are supported/

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+There are to 2 ways for cpma to obtain desired configurations:
+
+1. SSH, CPMA is capable to connect to cluster using SSH and fetch configurations to a local directory. For using SSH it's required to have a proper private key and SSH user may require sudo permissions for reading configurations.
```
gildub: I suggest something like: 
`1. Remotely retrieve configuration files using SSH, which are then stored locally (directory determined by '<workdir>' parameter) and under remote hostname sub directory. Proper SSH credentials (user, port and ssh key) are required with user having access to configuration files (sudo access might be necessary).`

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+2. Configurations stored in local directory, it's possible to manualy download configuration and feed them to CPMA.
```
gildub: Similarly:
`2. Locally stored configuration files. If option 1 is no possible, files can be retrieved prior to executing CPMA. The configuration files are expected to be located in '<workdir>' parameter and under the FQDN target hostname.`

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+---
+
+### 2. General CPMA configuration
+
+CPMA can be configured using either:
+
+1. Interactive prompts. This is simillar to `openshift-install`. The tool can be run with no configuration, all missing values will be prompted. You can see an example below. This is the most recommended way, because prompts will guide you through needed values and it can generate a configuration based on prompted values that will be used later.
```
gildub: Maybe clearer to say:
s/all missing values/all required values/

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+![prompt](https://user-images.githubusercontent.com/20123872/60581251-c0f57100-9d86-11e9-9ab3-7681b840731a.gif)
+
+
+2. Flags. All configuration values can be passed using flags. For example: `./cpma --source cluster.example.com --work-dir ./dir` Refer to CPMA's [README.md](https://github.com/fusor/cpma#usage) for full list of flags.
```
gildub: I suggest using CLI parameters instead of flags words.

docs/scenarios/manifests/README.md
gildub: Also I'd mention that the prompts allows to save the configuration which can be re-used at will.

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+3. Predefined configuration file. You can manually create a yaml configuration based on this [example](https://github.com/fusor/cpma/blob/master/examples/cpma-config.example.yaml).
+
+4. Environmental variables. It is also possible to pass all configuration values as environmental variables. List of variables can be found in [README.md](https://github.com/fusor/cpma#e2e-tests)
+
+---
+
+### 3. Using CPMA to generate manifests
+
+Once the configuration has been provided, either by prompt or by providing configuration file, manifests will be generated and placed in `output-directory/manifests`
```
gildub: I think we need to continue same train of thought as defined earlier:

`Once the configuration has been provided, by either prompt, CLI or ENV parameters, or configuration file,...`

docs/scenarios/manifests/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+If you have a predifined yaml configuration, you can pass it using --config flag, or place cpma.yaml in the directory from which you run cpma.
```
gildub: This needs to go to above `3. Predefined configuration file 'section'.

docs/README.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+### Report
+
+Report consist of 2 parts:
+
+1. Pre-migration analytics that goes under "cluster" in json. It contains general information about: 
+  * Nodes - name, сpu usage, memory capacity, consumed memory, number of running pods, pod capacity, information that
+  indicates if node is master.
+  * Namespaces - name, latest change timestamp, container count, total cpu/memory usage.
```
gildub: What about the items we list through namespaces ?

docs/README.md
```diff
@@ -0,0 +1,41 @@
...
+  * OAuth
+  * Project
+  * Scheduler
+  * SDN
+
+---
+
+### Manifests
+
+List of supported configuration to manifest translations:
```
gildub: Please add Cluster Quotas and Quotas Resources

docs/README.md
alexander-demichev: Right! Missed it

docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+  * Namespaces - name, latest change timestamp, container count, total cpu/memory usage.
+    * Pods - name
+    * Route - name, spec
+    * Daemonsets - name, latest change timestamp
+    * Deployments - name, latest change timestampg
+    * Quotas - name, spec, selectors(labels, annotations)
+  * Storage classes - name, provisioner
+  * RBAC - information about users, groups, roles, cluster roles, cluster role bindings, security context constraints
+  * Persistent volumes - names, storage class, driver, capacity, phase
+  * Quotas - name, spec, selectors(labels, annotations)
```
gildub: I kind of would like to use "Cluster Quotas" to clearly separate them from project (namespaces) quotas. What do you think?

docs/README.md
```diff
@@ -0,0 +1,47 @@
...
+  * Scheduler
+  * SDN
+
+---
+
+### Manifests
+
+List of supported configuration to manifest translations:
+  * Crio
+  * Cluster Quotas and Quotas Resources
```
gildub: Sorry I realize it's actually "Cluster Resources Quotas and Resources Quotas"

Issue #322, opened 2019-07-23
Configuration file: item not saved when it's missing


gildub:
If an item from configuration file is missing or added with new feature then survey would correctly prompt for the value if needed but the item is not saved to configuration file.

To easily address it, the configuration file choice (to save or not configuration) could be ask earlier in the survey process in order to save configuration if it's changed.

PR #323, opened 2019-07-23
Configuration file: Handle missing config items; more consistent types

gildub:
Allows to update configuration file if missing items.
Also makes all Boolean saved as such while survey still uses strings
Also makes port item a number (and still a string in survey).

Closes https://github.com/fusor/cpma/issues/322

### DIFF
--- a/pkg/transform/transform_test.go
+++ b/pkg/transform/transform_test.go
@@ -90,9 +90,6 @@ func TestOauthGenYAML(t *testing.T) {
 }
 
 func TestAllOtherCRGenYaml(t *testing.T) {
-	env.Config().Set("Manifests", true)
-	env.Config().Set("Reporting", true)
-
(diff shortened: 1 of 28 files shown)


pkg/transform/transform_test.go
```diff
@@ -15,6 +16,9 @@ import (
 )
 
 func TestOauthGenYAML(t *testing.T) {
+	env.Config().Set("Manifests", true)
+	env.Config().Set("Reporting", true)
```
Danil-Grigorev: Not sure if we need reports in such test.

gildub: Good catch, thanks

gildub: That said it fails without setting the reports. I'll leave for now.

PR #331, opened 2019-07-25
Test helpers: Renamed and Skipped

gildub:


pkg/transform/internal/test/load_test.go
```diff
@@ -0,0 +1,9 @@
+package test
+
+import (
+	"testing"
+)
+
+func TestTest(t *testing.T) {
+	t.Skip()
```
alexander-demichev: What was the reason for adding this test?

gildub: It shows that we care about test but this one doesn't have much coverage.

PR #343, opened 2019-08-09
Support for multiple output formats

alexander-demichev:
Since we are in feature freeze PR will go to a separate branch for now.
After some discussion, we decided that will be nice to support multiple report output formats: json, html, maybe pdf in future etc. 

This PR introduces 3 things: 

1. Improvement of writing report to json file.
Our current process of writing report to json file creates a json file before the transforms starts and then opens the file and appends a new report for every config part we have, this means the json file is opened, unmarshalled, marshalled a lot of times. 
I refactored this part, now all reports are collected and final report is passed to output function, report json is created only in case when report format is set to json and file is opened only once.

2. Adds support for multiple output formats.
I added new package `reportoutput`. This package is responsible for different output formats. It is scalable and has very simple structure.

3. Initial steps on HTML report.
`html.go` in `reportoutput` will be handling html report format. It is using go templates and inspired by go cover tool - https://github.com/golang/tools/blob/master/cmd/cover/html.go#L195

### DIFF
--- a/pkg/transform/reportoutput/json.go
+++ b/pkg/transform/reportoutput/json.go
@@ -1,0 +1,87 @@
+package reportoutput
+
+import (
+	"encoding/json"
+
+	"github.com/fusor/cpma/pkg/io"
+	"github.com/pkg/errors"
+)
+
+func jsonOutput(r ReportOutput) error {
+	var existingReports ReportOutput
+
+	jsonFile := "report.json"
+	emptyReport := []byte("{}")
+
+	if err := io.WriteFile(emptyReport, jsonFile); err != nil {
+		return errors.Wrapf(err, "unable to open report file: %s", jsonFile)
+	}
+
+	jsonData, err := io.ReadFile(jsonFile)
+	if err != nil {
+		return errors.Wrapf(err, "unable to read to report file: %s", jsonFile)
+	}
+
+	if err := json.Unmarshal(jsonData, &existingReports); err != nil {
+		return errors.Wrap(err, "unable to unmarshal existing report json")
+	}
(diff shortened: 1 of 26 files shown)


pkg/transform/reportoutput/json.go
```diff
@@ -0,0 +1,87 @@
...
+	"encoding/json"
+
+	"github.com/fusor/cpma/pkg/io"
+	"github.com/pkg/errors"
+)
+
+func jsonOutput(r ReportOutput) error {
+	var existingReports ReportOutput
+
+	jsonFile := "report.json"
```
alexander-demichev: This function code was copied from old `DumpReport()` function.  I see places for improvements here, just need some time for it.

pkg/transform/reportoutput/output.go
```diff
@@ -0,0 +1,45 @@
...
+	Kind       string `json:"kind"`
+	Supported  bool   `json:"supported"`
+	Confidence int    `json:"confidence"`
+	Comment    string `json:"comment"`
+}
+
+// DumpReports creates OCDs files
+func DumpReports(r ReportOutput) {
+	// reportOutputFormat := env.Config().GetString("OutputFormat")
+	reportOutputFormat := "all"
```
alexander-demichev: `reportOutputFormat` will be hardcoded to `all` before I add configurations for it.

alexander-demichev:
I added a big commit with CI fix. If PR is big for review at once, you can review it by commits.

alexander-demichev:
Closes https://github.com/fusor/cpma/pull/343

PR #346, opened 2019-08-13
Bundle html, css and js in binary, add new templates

alexander-demichev:
This PR introduces bundling HTML, CSS and JS files in binary. This is needed for cases when client machine will not have access to external network and scripts/styles can't be located remotely.

Bundling is done using https://github.com/GeertJohan/go.rice , a Go source file rice-box.go is created before build and it contains content of all assets. 

Lets merge the bundling and I will improve code from here.


.gitignore
```diff
@@ -30,3 +30,5 @@ coverage.*
 *.log
 
 cpma-config.yaml
+
+rice-box.go
```
jwmatthews: I think we will need this file to be included in source code in order to build this from release-engineering tools.

I doubt we will be successful in getting rel-eng processes to run 'rice embed-go' prior to building CPMA.

Pattern we've followed in past for these kind of things are to check in the generated resources.
It's not ideal, but it's been the path of least pain for integrating with processes.

jwmatthews: We decided to leave this in .gitignore for now, will revisit during rel-eng work and may end up committing this file.

PR #349, opened 2019-08-14
Add missing templates for cluster report

alexander-demichev:


jwmatthews:
Ran this and verified we have more info now under the Components, showing Red, Yellow, Green with confidence level.

LGTM

Issue #358, opened 2019-08-21
HTML report,  break down the namespaces with sub trees


alexander-demichev:
Break down the namespaces with sub trees in the same way as RBAC

PR #360, opened 2019-08-21
 Break down the namespaces with sub trees

alexander-demichev:
This PR: 
Breaks down the namespaces with sub trees.
For empty table cells sets text to "None".

Closes https://github.com/fusor/cpma/issues/358


alexander-demichev:
Screenshot 2019-08-21 at 14 07 18
Screenshot 2019-08-21 at 14 32 01

Issue #356, opened 2019-08-20
HTML Report, include PVs associated to a namespace


jwmatthews:
In the html report, would like to see info on PVs associated to a namespace.

Cluster Report -> Namespaces

would like to see another column capturing PVC/PV info.
For the info displayed, lets include
  PVC name, PV name, access mode, reclaim policy, capacity, storage class

PR #362, opened 2019-08-22
Add PVC to html/json report

alexander-demichev:
This PR adds PVC to html/json report
Screenshot 2019-08-22 at 13 54 09

Closes https://github.com/fusor/cpma/issues/356

### DIFF
--- a/pkg/transform/reportoutput/html.go
+++ b/pkg/transform/reportoutput/html.go
@@ -99,6 +99,15 @@ func parseTemplates() (*template.Template, error) {
 		"incrementIndex": func(i int) int {
 			return i + 1
 		},
+		"pvcIndex": func() *int {
+			i := new(int)
+			*i = 0
+			return i
+		},
+		"incrementPVCIndex": func(i *int) int {
+			*i = *i + 1
+			return *i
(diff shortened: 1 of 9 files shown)


jwmatthews:
cc @tsanders-rh the PV/PVC info for labs you requested

pkg/transform/reportoutput/html.go
```diff
@@ -99,6 +99,15 @@ func parseTemplates() (*template.Template, error) {
 		"incrementIndex": func(i int) int {
 			return i + 1
 		},
+		"pvcIndex": func() *int {
```
alexander-demichev: PVC index is special because it needs to be mutated. This index doesn't count items in one array, it counts PVCs that come from different arrays.

tsanders-rh:
Great.  Is "Name" as shown in the report "Namespace"?  Assume so.

alexander-demichev:
@tsanders-rh Updated screenshot
Screenshot 2019-08-22 at 17 01 53

tsanders-rh:
@alexander-demichev Perfect!  Thanks!

Issue #355, opened 2019-08-20
HTML report, sorting on columns


jwmatthews:
Would be nice if we could click on the headers of each column and select how data is sorted.
Assuming something simple at first of only sorting on one single column at a time

For example, looking at:
Cluster Report -> Namespaces

It would help if we could click on 'Latest Change' and have the rows sorted based on ascending/descending.

PR #367, opened 2019-08-23
Add table sorting

alexander-demichev:
This PR adds basic table sorting.
Screenshot 2019-08-23 at 16 44 21

Closes https://github.com/fusor/cpma/issues/355


pkg/transform/reportoutput/staticpage/js/custom.js
```diff
@@ -33,3 +40,82 @@ function fillEmptyTableData() {
...
+    var parentTable = $(this).parents()[2];
+    var parentTableHead = $(parentTable).children('thead');
+    var tableHeaders = $($(parentTableHead).children('tr')).children('th');
+
+    var parentTableBody = $(parentTable).children('tbody');
+    var tableRows = $(parentTableBody).children('tr').toArray();
+
+    var sorted = $.parseJSON($(this).attr('sorted'));
+
+    if (!sorted) {
```
gildub: Avoid code duplication here might be overkill but do we need the same comment twice?

alexander-demichev: Right, I thought this I removed it.

Issue #347, opened 2019-08-13
Improve warnings/error when unable the fetched contents of a remote file are incomplete.   Related to using 'root' as ssh user instead of 'ec2-user' in AWS OCP 3.x clusters


jwmatthews:
I saw the below warnings when running against an OCP 3.11 cluster deployed from
https://github.com/fusor/mig-agnosticd/tree/master/3.x

Master branch as of 8/13/2019, running as 'ec2-user'
```

$ ./cpma 
? Do you wish to save configuration for future use? true
? What will be the source for OCP3 config files? Remote host
? Path to crio config file /etc/crio/crio.conf
? Path to etcd config file /etc/etcd/etcd.conf
? Path to master config file /etc/origin/master/master-config.yaml
? Path to node config file /etc/origin/node/node-config.yaml
? Path to registries config file /etc/containers/registries.conf
? Do wish to find source cluster using KUBECONFIG or prompt it? KUBECONFIG
? Select cluster obtained from KUBECONFIG contexts master-jwm0-mg-example-com:443
? Select master node master1.example.internal
? SSH login root
? SSH Port 22
? Path to private SSH key /home/ec2-user/libra.pem
? Path to application data, skip to use current directory .
WARN[13 Aug 19 12:29 UTC] Skipping API - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
: open /home/ec2-user/.ssh/known_hosts: no such file or directory 
WARN[13 Aug 19 12:29 UTC] Skipping Crio - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
: open /home/ec2-user/.ssh/known_hosts: no such file or directory 
WARN[13 Aug 19 12:29 UTC] Skipping Docker - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
: open /home/ec2-user/.ssh/known_hosts: no such file or directory 
WARN[13 Aug 19 12:29 UTC] Skipping ETCD - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
: open /home/ec2-user/.ssh/known_hosts: no such file or directory 
WARN[13 Aug 19 12:29 UTC] Skipping OAuth - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
: open /home/ec2-user/.ssh/known_hosts: no such file or directory 
WARN[13 Aug 19 12:29 UTC] Skipping SDN - Unable to get hostkey in /home/ec2-user/.ssh/known_hosts
[...]

PR #378, opened 2019-08-28
Improve config decode error messages

alexander-demichev:


alexander-demichev:
Closes https://github.com/fusor/cpma/issues/347

pkg/transform/crio_transform.go
```diff
@@ -179,7 +179,7 @@ func (e CrioTransform) Extract() (Extraction, error) {
 
 	var extraction CrioExtraction
 	if _, err := toml.Decode(string(content), &extraction); err != nil {
-		return nil, errors.Wrap(err, "Failed to decode crio, see error")
+		return nil, errors.Wrap(err, "Failed to decode crio. Check ssh user permissions and file content, see error")
```
gildub: While at it, what about using `CRI-O` instead?

pkg/decode/decode.go
```diff
@@ -18,7 +18,7 @@ func MasterConfig(content []byte) (*legacyconfigv1.MasterConfig, error) {
 
 	serializer := k8sjson.NewYAMLSerializer(k8sjson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
 	if _, _, err := serializer.Decode(content, nil, masterConfig); err != nil {
-		return nil, errors.Wrap(err, "Failed to decode master config, see error")
+		return nil, errors.Wrap(err, "Failed to decode master config. Check ssh user permissions and file content, see error")
```
gildub: For this one and following `serializer.Decode` calls:
We should be able to detect a ssh error earlier -> connection/permission/etc/ issues and go that far in such cases.
So "Check file content" should suffice.

gildub:
This one is not needed anymore as https://github.com/fusor/cpma/issues/347 has been addressed.

PR #379, opened 2019-08-29
SSH fetchfile: check file exists first for better error mgt

gildub:
By checking if the file exists first allows for ssh session errors to not be caught into file retrieval step.


alexander-demichev:
LGTM

PR #387, opened 2019-09-06
Add vendor for downstream builds

jmontleon:


jmontleon:
OSBS has been firewalld off from the outside world for some months apparently. It wasn't always the case, but we now need everything vendored even to build images downstream.

PR #396, opened 2019-09-11
Add rpm spec

jmontleon:


jmontleon:
This is an RPM spec for building CPMA. It include a redistributable sub-package that contains binary builds for linux, mac, and windows.

Marked [DO NOT MERGE] until I have some time to review it better and maybe do some cleanup.

PR #397, opened 2019-09-11
Concurent api requests

gildub:
This is an initial approach to speed up API requests.
Early tests have shown a faster running time by a factor of 2.


gildub:
This PR has been replaced by https://github.com/fusor/cpma/pull/399
because of a reset of related fork/branch.

PR #406, opened 2019-09-18
Fix double encoding for oauth secrets

gildub:


arapov:
A ked, though the build issue must be addressed first.

PR #407, opened 2019-09-18
OAuth github: Filter out when organizations, teams are missing (#404)

gildub:


gildub:
Removing this one as `cherry-pick -x` was missing.

Issue #413, opened 2019-10-01
OAuth OpenId provider must warn about "issuer" 


gildub:
The OpenID provider field "issuer" is mandatory (no empty) for OCP4.
In OCP3, the are no such field (or equivalent) in the master configuration.

Can we draw the value from the existing configuration?
If not, then warn user to remember to enter a value for the "issuer" before applying the resource file.

gildub:
We could also use the following value "https://self-issued.me" providing self-signed ID Tokens in the meantime the user adjust the value.

PR #414, opened 2019-10-01
Report about OpenID issuer field 

gildub:
OCP4 OAuth OpenID provider has a new field dubbed "issuer" [1].
Meanwhile the field is mandatory and cannot be empty ("").
Using the "self-issued" value [2] allows creation. 

[1] https://docs.openshift.com/container-platform/4.1/authentication/identity_providers/configuring-oidc-identity-provider.html#identity-provider-oidc-CR_configuring-oidc-identity-provider
[2] https://openid.net/specs/openid-connect-core-1_0.html#SelfIssued

fixes https://github.com/fusor/cpma/issues/413
https://bugzilla.redhat.com/show_bug.cgi?id=1755865

### DIFF
--- a/pkg/transform/testdata/expected-report-oauth.json
+++ b/pkg/transform/testdata/expected-report-oauth.json
@@ -66,6 +66,13 @@
           "supported": true,
           "confidence": 2,
           "comment": "Identity provider my_openid_connect is supported in OCP4"
+        }, 
+        {
+          "name": "Issuer",
+          "kind": "IdentityProviders:OpenIDIdentityProvider",
+          "supported": true,
+          "confidence": 2,
+          "comment": "OCP4 requires an 'issuer' URL, please edit OAuth manifest file and configure this field"
         },
         {
           "name": "AccessTokenMaxAgeSeconds",
--- a/pkg/transform/oauth_transform.go
+++ b/pkg/transform/oauth_transform.go
@@ -121,6 +121,16 @@ func (e OAuthExtraction) buildReportOutput() {
 					Confidence: HighConfidence,
 					Comment:    fmt.Sprintf("Identity provider %s is supported in OCP4", p.Name),
 				})
+			if p.Kind == "OpenIDIdentityProvider" {
+				componentReport.Reports = append(componentReport.Reports,
+					reportoutput.Report{
+						Name:       "Issuer",
+						Kind:       fmt.Sprintf("IdentityProviders:%s", p.Kind),
+						Supported:  true,
+						Confidence: HighConfidence,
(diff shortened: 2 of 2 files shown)


arapov:
@gildub do you think that putting "https://self-issued.me" is safe? I mean it may be better to ask a person to fill this field in the resulting CR. 
Also, does ocp4 use this field in any functional way, e.g. trying to connect/verify/whatever with the issuer?

gildub:
@arapov, sorry I was updating the description with more details in the meantime.
The field is mandatory but we can certainly feed it a different value.

gildub:
@arapov, I can see unknown security implications. 
So let's use an example dummy value for now and at same time provide message to user when OpenID is handled.

pkg/transform/testdata/expected-report-oauth.json
```diff
@@ -67,6 +67,13 @@
           "confidence": 2,
           "comment": "Identity provider my_openid_connect is supported in OCP4"
         },
+        {
+          "name": "Issuer",
+          "kind": "IdentityProviders:OpenIDIdentityProvider",
+          "supported": true,
+          "confidence": 2,
+          "comment": "OCP4 requires an 'issuer' URL, please edit OAuth manifest file and configure this field"
```
arapov: @gildub do we have other places in the code where we require an action from user before the generated CRs could be used? I wonder, whether we want to stress even more that there are actions that are required to be fulfilled. WDYT?

The code looks Okay-ish to me though.

gildub: @arapov, we can survey user at any time but do we really want to introduce such interaction during execution of the tool?  I suppose that's a choice to make.

@jwmatthews what do you think?

jwmatthews: I would refrain from survey at this point in time, OK to require the user to review and modify the CR before being applied.

I think the note in the report that 'issuer' needs to be set is helpful and sufficient.

It's OK to have the CR have an empty field for the issuer and have it fail to be applied as described in
https://bugzilla.redhat.com/show_bug.cgi?id=1755865

I assume using 'https://no-issuer.me" could be confusing and hide the error, so would remove 'issuer: https://no-issuer.me' and go back to issuer: ""

gildub: @jwmatthews, thanks for your input, and helping us make a decision.

Let's reverse the default value to "" and keep the report message then.

arapov: +1 for the message and failing on import to ocp4.

gildub: @jwmatthews, thanks for your input, and helping us make a decision.

Let's reverse the default value to "" and keep the report message then.

PR #434, opened 2019-11-26
API requests to use client as parameter

gildub:
Allows API requests to be client independent


gildub:
This PR has been replaced with https://github.com/fusor/cpma/pull/436

PR #459, opened 2020-02-25
Renames user1 import to userv1

gildub:
This has been applied to master too but as part of patch which didn't make it to this branch


gildub:
Wrong branch!

PR #461, opened 2020-02-25
Removes unused target

gildub:


gildub:
Closing after konveyor move due to non-existing source branch.

PR #474, opened 2020-05-05
Update vendor for downstream

dymurray:


djwhatle:
@dymurray it looks like this also updates the content of the `go.mod` file. Is this intentional? I wonder if there's a way for us to pull deps into vendor without changing this? Maybe it's harmless but seems like a dependency change (more than just pulling into vendor), not sure if you intended to do this.

go.mod
```diff
@@ -25,23 +18,18 @@ require (
...
 	golang.org/x/time v0.0.0-20190308202827-9d24e82272b4 // indirect
 	google.golang.org/appengine v1.5.0 // indirect
-	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/go-ini/ini.v1 v1.42.0
-	gopkg.in/inf.v0 v0.9.0 // indirect
 	gopkg.in/ini.v1 v1.42.0 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
-	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
-	k8s.io/apimachinery v0.0.0-20190313205120-d7deff9243b1
+	k8s.io/api v0.18.2
```
djwhatle, line 29: Wondering why this jumped from a dated commit release to v0.18.2, did you change this on purpose @dymurray ?

dymurray:
@djwhatle I had to update go.mod because I got this error at first:

```
[dymurray@town cpma]$ go mod vendor
go: github.com/openshift/api@v3.9.1-0.20190404192821-4706c46ddae5+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```

So I think go.mod isn't correct right now. I'm also wondering why we are using 3.9 deps

djwhatle:
If we are upgrading the OpenShift / k8s API dependencies I think it's going to be a necessity to make sure CPMA can still communicate with older OCP versions. This tool is supposed to work with OpenShift 3.7+, so I'd probably just check that CPMA can be run against an OpenShift 3.7 cluster.

I think I have an OpenShift 3.7 cluster, so I can try to help with this test.

djwhatle (commented):
Want to test that the OpenShift API version upgrade doesn't break OCP 3.7 compatibility before we merge this.

dymurray:
This PR is currently broken I just don't know what to do. @gildub please help me get vendor updated with the latest dependencies so we can build downstream.

dymurray:
We turned Go modules off to fix this downstream. Let's please work to keep vendor up to date to prevent this from happening again.

